PR #1, opened 2019-05-07
Update wasm32.rs to name the union fields.

sunfishcode:
The upstream wasi/core.h header switched from anonymous unions to unions
named `u`. This patch updates wasm32.rs to reflect this, as was done in
wasmtime 5b77f952.


kubkon:
LGTM!

PR #4, opened 2019-05-07
Port WASI host types

kubkon:
Completely removes `bindgen` dependency, and majority of primitive type casts.


sunfishcode (approved):
I like this approach. Using bindgen to generate host.rs made sense initially, especially while many things were changing, but now the API has settled down. There will still be changes, but I expect they'll be manageable. This change removes a dependency on wasi-sdk, bindgen, etc., which is great since we've had quite a few reports of those causing trouble, and removing all those `at` casts is nice :-).

kubkon:
Now, it's also much easier to set up the CI which I'll do just now ;-)

PR #6, opened 2019-05-07
Set up CI, add format and test scripts.

kubkon:
Sets up CI, and adds format and test scripts. CI's setup and the scripts borrowed from [CraneStation/wasmtime](https://github.com/CraneStation/wasmtime) project. Also removes unused Cargo dependencies and fixes a trivial numeric cast.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -678,10 +678,17 @@ pub fn nix_from_filetype(sflags: __wasi_filetype_t) -> nix::sys::stat::SFlag {
 }
 
 pub fn filestat_from_nix(filestat: nix::sys::stat::FileStat) -> __wasi_filestat_t {
+    use std::convert::TryFrom;
+
     let filetype = nix::sys::stat::SFlag::from_bits_truncate(filestat.st_mode);
+    let dev = __wasi_device_t::try_from(filestat.st_dev)
+        .expect("FileStat::st_dev is trivially convertible to __wasi_device_t");
+    let ino = __wasi_inode_t::try_from(filestat.st_ino)
+        .expect("FileStat::st_ino is trivially convertible to __wasi_inode_t");
+
     __wasi_filestat_t {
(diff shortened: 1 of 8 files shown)


sunfishcode (commented):
This should be in Wasmtime too, but it'd be good to add this to the .rustfmt.toml file:
```
# This file tells tools we use rustfmt. We use the default settings.
```
so that we document the intent.

kubkon:
Done. I'll backport it to wasmtime too.

kubkon:
Also, I've had to move `trivial_numeric_casts` from `#deny` to `#warn` as this can lead to tricky behaviour depending on the OS we're compiling on (`libc::stat::FileStat::st_ino` resolving to a type different than `u64` on mac, but to the same on linux).

src/host.rs
```diff
@@ -680,6 +680,8 @@ pub fn nix_from_filetype(sflags: __wasi_filetype_t) -> nix::sys::stat::SFlag {
 pub fn filestat_from_nix(filestat: nix::sys::stat::FileStat) -> __wasi_filestat_t {
     let filetype = nix::sys::stat::SFlag::from_bits_truncate(filestat.st_mode);
     __wasi_filestat_t {
+        // TODO: with #deny on trivial numeric casts, the line below will pass
+        // on macos, but fail on linux
```
sunfishcode: Would it work to write this as `__wasi_device_t::from(filestat.st_dev)` and `__wasi_inode_t::from(filestat.st_ino)`?

kubkon: Unfortunately, it won't be portable as on my mac, for instance, `filestat.st_dev` is `i32` and `__wasi_device_t` being `u64` triggers trait bound not satisfied error for `u64: std::convert::From<i32>`.

sunfishcode: Ah, ok. And I see you're now using `try_from` here, which looks good.

sunfishcode:
Looks good!

PR #8, opened 2019-05-09
add .gitignore

afinch7:


.gitignore
```diff
@@ -0,0 +1,3 @@
+target
```
kubkon: If you could change that to `/target` to be compatible with Cargo defaults, that'd be great.

.gitignore
```diff
@@ -0,0 +1,3 @@
...
+Cargo.lock
+.*.rustfmt
```
kubkon: I'm not sure that's actually needed. However, instead, if you could add `**/*.rs.bk` to be compatible with Cargo defaults, that'd be awesome!

kubkon:
Hey! Thanks for this! It looks good, just a couple of minor nitpicking.

kubkon:
Looks good, thanks! Just one more thing, could you explain why you've added `.*.rustfmt`? I can't seem to be able to find a use case for it with my setup.

afinch7:
I was looking at the .gitignore in wasmtime for reference. I wasn't sure why it was there either, but it looks like it was added a long time ago in wasmtime.

kubkon:
Ah, ok, good to know! I'm not sure why it's there either. Perhaps @sunfishcode could shed some light on this?

sunfishcode:
I'm not sure why it's there either, offhand. Removing it is fine with me!

kubkon:
LGTM, thanks!

PR #9, opened 2019-05-09
Add contributing guide and code of conduct

kubkon:


sunfishcode:
Great! For people following along, this is the same code of conduct that's used in Cranelift and Wasmtime.

PR #10, opened 2019-05-10
Add experimental struct with views into memory and ctx

kubkon:
This PR adds experimental struct `VmContextView` with (mutable) views into guest's memory and (WASI) context. This struct is heavily taking from Lucet's [Vmctx struct](https://github.com/fastly/lucet/blob/master/lucet-runtime/lucet-runtime-internals/src/vmctx.rs#L21).

I thought I'd stimulate the discussion about the potential API with a proposition, and here it is. The changes introduced in this PR reflect directly upon my in-depth offline discussions with @sunfishcode and pointers from @acfoltzer (see #7 for some previous context).

The `VmContextView` struct is essentially mimicking the Lucet's `Vmctx` struct, however, directed more towards WASI use. It is structured as follows:

```rust
pub struct VmContextView {
    pub memory_view: RefCell<Box<[u8]>>,
    pub wasi_ctx_view: RefCell<Box<WasiCtx>>,
}
```

The `RefCell`s here serve the purpose mentioned by @acfoltzer: they circumvent potential gross aliasing in the implementation of the hostcalls. Note that in this prototype, I didn't consider memory growth at all, in the sense that, if the memory was grown in-between the calls, the `memory_view` will become invalid. In Lucet, this is nicely handled using a [reconstitute_heap_view_if_needed](https://github.com/fastly/lucet/blob/master/lucet-runtime/lucet-runtime-internals/src/vmctx.rs#L134) routine, which could potentially be added here as well in some way, should we agree that this PR is the road we want to take.

The `VmContextView` struct can then be created via `AsVmContextView` trait:

```rust
pub trait AsVmContextView {
    unsafe fn as_vm_context_view(&mut self) -> VmContextView;
}
```

The trait fn is unsafe as the `VmContextView` does indeed allow for mutable access to the guest's memory and WASI context from behind `&-ref` at runtime. I wasn't sure whether a trait like this is the way to go, but I'm hopeful that with your help guys, we can come up with a nice solution.

I should note here that I've tested the changes in Wasmtime, and it does work with this approach (see [here](https://github.com/kubkon/wasmtime/blob/wasi-common-api/wasmtime-wasi/src/syscalls.rs#L14)).

EDIT: forgot to mention that, as a result of this change, the hostcalls are now (almost) 1:1 with [lucet-wasi](https://github.com/fastly/lucet/blob/master/lucet-wasi/src/hostcalls.rs)'s with the exception of `proc_exit` due to custom `panic` unwinding in Lucet, but that can be handled later IMHO.


src/ctx.rs
```diff
@@ -198,6 +198,16 @@ pub struct WasiCtx {
     pub env: Vec<CString>,
 }
 
+impl Default for WasiCtx {
```
kubkon: This is just temporary, but I needed it to properly handle `drop` for `VmContextView`, basically following the way it's done in Lucet.

src/vmctx.rs
```diff
@@ -0,0 +1,59 @@
...
+        Box::leak(wasi_ctx_view);
+    }
+}
+
+impl VmContextView {
+    pub fn memory(&self) -> Ref<[u8]> {
+        let r = self
+            .memory_view
+            .try_borrow()
+            .expect("memory not already borrowed mutably");
```
kubkon: The `panic`s here could be handled in a different way, perhaps even using the same mechanism for `panic` unwinding that is currently used in Lucet (see [hostcall_macros.rs](https://github.com/fastly/lucet/blob/master/lucet-runtime/lucet-runtime-internals/src/hostcall_macros.rs#L48)).

src/memory.rs
```diff
@@ -23,34 +23,55 @@ macro_rules! bail_errno {
     };
 }
 
+pub unsafe fn dec_ptr(
```
kubkon, line 25: Since `VmContextView` has a view at the guest's memory now, it was possible to bring this function back in full.

sunfishcode (commented):
If I understand correctly, this is used `Box` to hold non-owned data, by clearing out the `Box` in the `drop` function that contains it. This is pretty tricky :-}.

I wonder if it's possible to insulate `wasi-common` from the tricky parts even more. Would it work to have signatures like this?
```rust
pub fn fd_read(
    wasi_ctx: &mut WasiCtx,
    memory: &mut [u8],
    fd: wasm32::__wasi_fd_t,
    iovs_ptr: wasm32::uintptr_t,
    iovs_len: wasm32::size_t,
    nread: wasm32::uintptr_t,
) -> wasm32::__wasi_errno_t
```
And then put the responsibility for obtaining the `wasi_ctx` and `memory` &mut references on the wrapper code. For some system calls, we could make those `&` instead of `&mut`. And for some like `sched_yield`, we could leave out the `memory` argument altogether.

It seems like this should be safe to do, because WASI functions don't resize memory, and memory doesn't alias the `WasiCtx`. And that way, if someone is calling into wasi-common from plain Rust, they can do so through an entirely safe API.

We'll still need a way to obtain those `&mut` references, but it seems desirable to keep that separate from the core WASI implementation logic. Does that sound reasonable?

kubkon:
Yep, you're absolutely right about `RefCell<Box<T>>`. This brilliant trick is AFAIK by @acfoltzer, and I've borrowed it from Lucet implementation of the VM context struct :-)

I actually really like the idea of insulating the API, and pushing the trickery back to wrapper code. So +1 from me.

I've pushed the suggested changes as a separate commit in this PR so that we can easily track progress of our discussion, but after we agree on something concrete with the intention to merge the PR, I'll be more than happy to clean the PR up with rebases and/or squashes.

Furthermore, I believe that the changes you've suggested @sunfishcode should work well in both wasmtime and lucet. In fact, I've already verified it in wasmtime and works just fine. If time permits, I'll also check lucet later today.

Finally, to sum, I'm voting in favour of @sunfishcode suggested changes, and would very much like to see them merged :-)

sunfishcode:
Cool, this looks good.

Let's merge this now and work from here, but while here I did notice another opportunity -- I think we can make `dec_ptr_to` a safe function now, by adding a lifetime parameter like this:
```rust
pub unsafe fn dec_ptr_to<T, 'memory>(
    vmctx: *mut VmContext,
    memory: &'memory mut [u8],
    ptr: wasm32::uintptr_t,
) -> Result<&'memory mut T, host::__wasi_errno_t> {
```
It'll still need an `unsafe` block internally to form the `&mut`, but it can check size and alignment and present a safe interface to its users.

Similary, we could make `dec_slice_of` a safe function which returns a `&'memory mut [T]` rather than a pointer and length pair. And if we do those, some WASI implementation functions could switch from `unsafe fn` to `fn` :-).

kubkon:
Ah, good one! I'll add these in a separate PR :+1:

PR #11, opened 2019-05-12
Clean up docs; specify minimum Rust version

kubkon:
Summary of changes:
* set Rust version to 1.34+; required due to the use of `std::convert::TryFrom` which wasn't available in earlier versions
* remove obsolete license files
* clean up README, and update the list of all the implemented hostcalls in the lib


sunfishcode (approved):
Looks good!

PR #12, opened 2019-05-12
Make memory fns safe wherever possible

kubkon:
As suggested in #10, I've reworked parts of `memory.rs` so that the vast majority of the enc/dec fns are now safe. This took away quite a few `unsafe` calls in the hostcall impl code. @sunfishcode lemme know if this is what you've had in mind.

While at it, I've also done some minor refactoring to some hostcall impl functions.

### DIFF
--- a/src/memory.rs
+++ b/src/memory.rs
@@ -14,6 +14,7 @@ use crate::{host, wasm32};
 use cast;
 use cast::From as _0;
 use std::mem::{align_of, size_of};
+use std::ptr;
 use std::slice;
 
 macro_rules! bail_errno {
@@ -22,56 +23,53 @@ macro_rules! bail_errno {
     };
 }
 
-pub unsafe fn dec_ptr(
+pub fn dec_ptr(
     memory: &mut [u8],
     ptr: wasm32::uintptr_t,
     len: usize,
 ) -> Result<*mut u8, host::__wasi_errno_t> {
-    // check that `len` fits in the wasm32 address space
-    if len > wasm32::UINTPTR_MAX as usize {
-        bail_errno!(__WASI_EOVERFLOW);
-    }
-
-    // check that `ptr` and `ptr + len` are both within the guest heap
-    if ptr as usize > memory.len() || ptr as usize + len > memory.len() {
-        bail_errno!(__WASI_EFAULT);
-    }
+    // check for overflow
+    let checked_len = (ptr as usize).checked_add(len).ok_or(host::__WASI_EFAULT)?;
 
     // translate the pointer
-    Ok(memory.as_mut_ptr().offset(ptr as isize))
+    memory
+        .get_mut(ptr as usize..checked_len)
+        .ok_or(host::__WASI_EFAULT)
+        .map(|mem| mem.as_mut_ptr())
 }
 
-pub unsafe fn dec_ptr_to<T>(
-    memory: &mut [u8],
+pub fn dec_ptr_to<'memory, T>(
+    memory: &'memory mut [u8],
     ptr: wasm32::uintptr_t,
-) -> Result<*mut T, host::__wasi_errno_t> {
+) -> Result<&'memory mut T, host::__wasi_errno_t> {
     // check that the ptr is aligned
     if ptr as usize % align_of::<T>() != 0 {
         bail_errno!(__WASI_EINVAL);
     }
-    dec_ptr(memory, ptr, size_of::<T>()).map(|p| p as *mut T)
+
+    dec_ptr(memory, ptr, size_of::<T>()).map(|p| unsafe { &mut *(p as *mut T) })
 }
 
-pub unsafe fn dec_pointee<T>(
+pub fn dec_pointee<T>(
     memory: &mut [u8],
     ptr: wasm32::uintptr_t,
 ) -> Result<T, host::__wasi_errno_t> {
-    dec_ptr_to::<T>(memory, ptr).map(|p| p.read())
+    dec_ptr_to::<T>(memory, ptr).map(|p| unsafe { ptr::read(p) })
 }
 
-pub unsafe fn enc_pointee<T>(
+pub fn enc_pointee<T>(
     memory: &mut [u8],
     ptr: wasm32::uintptr_t,
     t: T,
 ) -> Result<(), host::__wasi_errno_t> {
-    dec_ptr_to::<T>(memory, ptr).map(|p| p.write(t))
+    dec_ptr_to::<T>(memory, ptr).map(|p| unsafe { ptr::write(p, t) })
 }
 
-pub unsafe fn dec_slice_of<T>(
-    memory: &mut [u8],
+pub fn dec_slice_of<'memory, T>(
+    memory: &'memory mut [u8],
     ptr: wasm32::uintptr_t,
     len: wasm32::size_t,
-) -> Result<(*mut T, usize), host::__wasi_errno_t> {
+) -> Result<(&'memory mut T, usize), host::__wasi_errno_t> {
     // check alignment, and that length doesn't overflow
     if ptr as usize % align_of::<T>() != 0 {
         return Err(host::__WASI_EINVAL);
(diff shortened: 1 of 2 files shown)


sunfishcode:
Nice!

Can we also make `dec_ptr` safe? Instead of `memory.offset(ptr)`, we could use `memory.get(ptr, ptr + len)`, and if that succeeds, `.as_mut_ptr()` on the resulting slice.

I'd find it more consistent for `dec_ptr` to return `EFAULT` rather than `EOVERFLOW` in the case where `len` is greater than `wasm32::UINTPTR_MAX` -- `EOVERFLOW` is mostly used in cases where values being returned to the user don't fit in the user-provided data types, and in this case, we're not returning data back to the user. Using `EFAULT` here also means we could eliminate the explicit bounds checks -- the `get` call will check everything and you can just return `EFAULT` if it fails.

As one more detail while here, the `ptr + len` should be a `checked_add` to handle overflow there as well (and that should also fail with `EFAULT` for similar reasons.)

kubkon:
Done. I have to admit that I wasn't sure about the signature of `dec_ptr`, i.e., whether it should be safe or unsafe as it returns a raw pointer. But I've convinced myself that it's valid and consistent with Rust's stdlib (e.g., `std::slice::as_mut_ptr` is marked safe even though it returns a raw pointer as well).

I've also changed `EOVERFLOW` to `EFAULT`, and essentially ported over `checked_add` logic from `wasmtime-wasi`. Question here though: should we worry about `len == 0` in `dec_ptr`?

sunfishcode:
> Done. I have to admit that I wasn't sure about the signature of `dec_ptr`, i.e., whether it should be safe or unsafe as it returns a raw pointer. But I've convinced myself that it's valid and consistent with Rust's stdlib (e.g., `std::slice::as_mut_ptr` is marked safe even though it returns a raw pointer as well).

Yeah, from a language design point of view, Rust's choice here is debatable, but given that it works this way, it makes sense for us to do the same thing.

> I've also changed `EOVERFLOW` to `EFAULT`, and essentially ported over `checked_add` logic from `wasmtime-wasi`. Question here though: should we worry about `len == 0` in `dec_ptr`?

That's a great question :-). I think the main interesting place where this would come up in wasm is in the [bulk memory operations](https://github.com/WebAssembly/bulk-memory-operations/). Looking into it, it appears that the current proposal isn't clear on what happens with an empty slice at the end of memory, so I filed https://github.com/WebAssembly/bulk-memory-operations/pull/86. On the Rust side, it isn't clear from the documentation what `get` does on an empty slice at the end of a slice either, so I filed https://github.com/rust-lang/rust/issues/60783.

It appears the intent of both WebAssembly linear memory and Rust have the behavior that a zero length is ok, including at the end of memory/slice, but not past the end of memory/slice, so we can just rely on `get`'s bounds checking.

kubkon:
> > Done. I have to admit that I wasn't sure about the signature of `dec_ptr`, i.e., whether it should be safe or unsafe as it returns a raw pointer. But I've convinced myself that it's valid and consistent with Rust's stdlib (e.g., `std::slice::as_mut_ptr` is marked safe even though it returns a raw pointer as well).
> 
> Yeah, from a language design point of view, Rust's choice here is debatable, but given that it works this way, it makes sense for us to do the same thing.

Agreed, although in Rust's defence, you still need an `unsafe` block to dereference the raw pointer anyway, so it might not be as bad as it looks ;-)

> 
> > I've also changed `EOVERFLOW` to `EFAULT`, and essentially ported over `checked_add` logic from `wasmtime-wasi`. Question here though: should we worry about `len == 0` in `dec_ptr`?
> 
> That's a great question :-). I think the main interesting place where this would come up in wasm is in the [bulk memory operations](https://github.com/WebAssembly/bulk-memory-operations/). Looking into it, it appears that the current proposal isn't clear on what happens with an empty slice at the end of memory, so I filed [WebAssembly/bulk-memory-operations#86](https://github.com/WebAssembly/bulk-memory-operations/pull/86). On the Rust side, it isn't clear from the documentation what `get` does on an empty slice at the end of a slice either, so I filed [rust-lang/rust#60783](https://github.com/rust-lang/rust/issues/60783).
> 
> It appears the intent of both WebAssembly linear memory and Rust have the behavior that a zero length is ok, including at the end of memory/slice, but not past the end of memory/slice, so we can just rely on `get`'s bounds checking.

Oh cool! Thanks for clarifying it and filing the PR and the issue!

Unless you have any other observations, I reckon the PR looks pretty good now, and I'm happy we managed to ged rid of a fair chunk of unsafe-{ty, ness} from the code. :-)

src/memory.rs
```diff
@@ -22,56 +23,62 @@ macro_rules! bail_errno {
...
+pub fn dec_ptr(
     memory: &mut [u8],
     ptr: wasm32::uintptr_t,
     len: usize,
 ) -> Result<*mut u8, host::__wasi_errno_t> {
     // check that `len` fits in the wasm32 address space
     if len > wasm32::UINTPTR_MAX as usize {
-        bail_errno!(__WASI_EOVERFLOW);
+        bail_errno!(__WASI_EFAULT);
     }
```
sunfishcode: This check turns out to be redundant, because a wasm32 program will never have a memory larger than `wasm32::UINTPTR_MAX`, so `ptr + len` will simply be out of bounds if `len` is too great.

src/memory.rs
```diff
@@ -22,56 +23,62 @@ macro_rules! bail_errno {
...
 
-    // check that `ptr` and `ptr + len` are both within the guest heap
-    if ptr as usize > memory.len() || ptr as usize + len > memory.len() {
-        bail_errno!(__WASI_EFAULT);
+    // check for overflow
+    if len > 0 {
+        (ptr as usize)
+            .checked_add(len - 1)
+            .ok_or(host::__WASI_EFAULT)?;
     }
```
sunfishcode: This checks for overflow with `ptr + (len - 1)`, but below we actually compute `ptr + len`, so currently this overflow check doesn't perfectly match the code below. I don't think it's an observable difference in this case, but I'd still like to straighten this out, as this is in foundational code we'll build a lot of other things on :-).

The `- 1` handles the case where the subrange is at the very end of the index space, however this add is happening under the host `usize` index space, so the only way `ptr + len` could overflow where `ptr + (len - 1)` couldn't, without getting caught by the bounds check, is if `memory` is the same size as the entire host address space, and that's not possible :-).

So first suggestion: do the `checked_add` without the `- 1`. Then on the success path, save the `checked_add` result in a local variable and use that below, rather than doing a separate unchecked add, as that makes it more obvious that the actual value used below is the checked one.

(In theory, an exotic host could have a `usize` which isn't big enough to index the entire address space, and allow objects larger than `isize::max_value()` which is [not trivial to do](https://github.com/rust-lang/unsafe-code-guidelines/issues/102). If anyone ports this code to such a platform, the change I suggest here would cause the code to return `EFAULT` in a case where it technically shouldn't, so it'll "fail closed", which I posit is adequate for now.

Another option would be to use an inclusive range with a `- 1` like `memory.get_mut(ptr as usize..=ptr as usize + (len - 1))`, however that would require special-casing when `len` is 0, and my sense is this issue is not worth that complexity.)

sunfishcode (commented):
Thanks for your patience here! My sense here is that having a simple and safe `dec_ptr` will be worth the effort :-).

kubkon:
> Thanks for your patience here! My sense here is that having a simple and safe `dec_ptr` will be worth the effort :-).

Totally agree here, so no need to thank me -- always up for a clean and safe implementation especially when it comes to something as mission-critical as memory management! :-)

src/memory.rs
kubkon: Ah, good point!

src/memory.rs
kubkon: Good catch about the bounds check here. I should have noticed it but somehow slipped through! ;-)

Personally I also prefer the first suggestion over the second one. IMHO, less nesting means better code and most importantly readability.

sunfishcode:
Cool, I'm excited about all these functions switching from unsafe to safe! Of course there still are some unsafe parts, but they're much easier to see and understand now.

PR #13, opened 2019-05-13
Make users of `dec_slice_of` safe.

sunfishcode:
Make `dec_slice_of` return a slice rather than a pointer-length pair,
freeing its users from having to call the unsafe `slice::from_raw_parts`.

This requires splitting `dec_slice_of` and `dec_ptr` into mut and
non-mut versions, and reorganizing poll_oneoff a little to avoid
borrow-checker errors -- decoded slices do alias the main memory, so
make sure functions only need one or the other.

### DIFF
--- a/src/hostcalls.rs
+++ b/src/hostcalls.rs
@@ -542,13 +542,12 @@ pub fn path_open(
     }
 
     let path = match dec_slice_of::<u8>(memory, path_ptr, path_len) {
-        Ok((ptr, len)) => OsStr::from_bytes(unsafe { std::slice::from_raw_parts(ptr, len) }),
+        Ok(slice) => OsStr::from_bytes(slice),
         Err(e) => return enc_errno(e),
     };
 
     let (dir, path) = match path_get(
         wasi_ctx,
-        memory,
         dirfd,
         dirflags,
         path,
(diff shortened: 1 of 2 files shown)


sunfishcode:
This will conflict with https://github.com/fastly/lucet/pull/176 -- it moves some of the code around, and contains more `dec_slice_of` callsites that this PR will enable to be safe, so feel free to wait on reviewing this if you want, and I'll merge it once the other PR lands.

src/hostcalls.rs
```diff
@@ -1159,24 +1154,18 @@ fn poll_oneoff_handle_timeout_event(
             },
         };
         output_slice[0] = enc_event(output_event);
-        if let Err(e) = enc_pointee(memory, nevents, 1) {
```
kubkon, line 1164: So, is `enc_pointee` no longer needed here?

src/hostcalls.rs
```diff
@@ -1248,18 +1237,14 @@ fn poll_oneoff_handle_fd_event<'t>(
         *output_slice_cur.next().unwrap() = enc_event(output_event);
         revents_count += 1;
     }
-    if let Err(e) = enc_pointee(memory, nevents, revents_count) {
```
kubkon, line 1253: Like above, is `enc_pointee` obsolete in this hostcall?

kubkon:
Looks great to me! I like the fact we now have `&[u8]` in decode fns. :-)

In terms of merging it now or after fastly/lucet#176 lands, I’m happy either way. If need be, I’m happy to make an effort in adapting it to this repo.

kubkon (approved):

src/hostcalls.rs
sunfishcode: The call to `enc_pointee` moved up a level, into the body of `poll_oneoff`. That avoids the borrow-checker error with `memory` aliasing `output_slice` within the body of `poll_oneoff_handle_timeout_event`. It's also a little tidier, because we only need one call to enc_pointee` at the end of `poll_oneoff` rather than three in total, on the various paths in `poll_oneoff_handle_timeout_event` and `poll_oneoff_handle_fd_event`.

src/hostcalls.rs
sunfishcode: See above :-).

sunfishcode:
I'm inclined to land this, since we already landed 8b09f321ac230d5ca9d33987c90c76395d44709b and a few other things, so there's already some merging to do. If you want to merge 8b09f321ac230d5ca9d33987c90c76395d44709b when it lands, that'd be great, although feel free to ask me to help with it too :-).

kubkon:
Excellent point! :-)

PR #15, opened 2019-05-14
Export C bindings to hostcalls

kubkon:
This PR adds C bindings to the hostcall functions. This is achieved through the use of `proc_macro_attribute` currently called `#[wasi_common_cbindgen]` but this up for debate. The use of proc macro is largely inspired by [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen); however, it is currently only tailored for our very restricted use case, and also more of a prototype than anything else. Nonetheless, it manages to generate correct C bindings just fine.

To provide an illustrative example, the hostcall annotated with `#[wasi_common_cbindgen]`

```rust
#[wasi_common_cbindgen]
pub fn args_get(
    wasi_ctx: &WasiCtx,
    memory: &mut [u8],
    argv_ptr: wasm32::uintptr_t,
    argv_buf: wasm32::uintptr_t,
) -> wasm32::__wasi_errno_t {
    // ...
}
```

expands to the original, annotated function untouched plus a C binding of the form

```rust
#[no_mangle]
pub unsafe extern "C" fn __wasi_args_get(
    wasi_ctx: *mut WasiCtx,
    memory: *mut u8,
    memory_len: usize,
    argv_ptr: wasm32::uintptr_t,
    argv_buf: wasm32::uintptr_t,
) -> wasm32::__wasi_errno_t {
    args_get(
        &mut *wasi_ctx,
        std::slice::from_raw_parts_mut(memory, memory_len),
        argv_ptr,
        argv_buf,
    )
}

```

A point in favour of this approach is the fact that it works for arbitrary hostcall signatures which would be difficult (if not impossible) to achieve with `macro_rules!` macros. If we decide that we're happy with this as the general direction for generating C bindings, I will port the mechanism back to wasmtime, and hopefully clean up the implementation of syscalls (see CraneStation/wasmtime#138 for reference).

In addition, I've also added a series of very simple, minimalistic tests to the proc macro, which should be executed automatically by the CI.

NB you can inspect the generated C bindings with `cargo expand` ;-)


kubkon:
In the first iteration, I've decided to go the lazy way and for every `&` or `&mut` just stick in `*mut`, but it could definitely be changed to actually recognise mutability and substitute `*const` for `&` and `*mut` for `&mut`.

wasi-common-cbindgen/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+authors = ["Jakub Konka <kubkon@jakubkonka.com>"]
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = { version = "0.15.34", features = ["full"] }
+quote = "0.6.12"
+trybuild = "1.0.4"
```
sunfishcode, line 15: trybuild is only used for tests; can it be a dev-dependency?

wasi-common-cbindgen/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+name = "wasi-common-cbindgen"
+version = "0.1.0"
+authors = ["Jakub Konka <kubkon@jakubkonka.com>"]
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = { version = "0.15.34", features = ["full"] }
```
sunfishcode, line 11: If I read this right, this adds syn as a transitive dependency of wasi-common. Does that add to wasi-common's own footprint at all? Or does the fact that this is a proc-macro crate mean that it's just a compile-time dependency?

wasi-common-cbindgen/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+    assert!(attr.is_empty());
+
+    let function = syn::parse_macro_input!(function as syn::ItemFn);
+
+    // capture visibility
+    let vis = &function.vis;
+
+    // generate C fn name prefixed with __wasi_
+    let fn_ident = &function.ident;
+    let concatenated = format!("__wasi_{}", fn_ident);
```
sunfishcode: WASI libc uses `__wasi_` because double-underscore prefixes are reserved for libc/compiler use. And since the interfaces generated here don't exactly match `<wasi/core.h>` -- they have explicit vmctx and memory arguments -- would it make sense to change the prefix to `wasi_` or `wasi_common_` or so?

sunfishcode (approved):
Wow, this is really cool.

It is possible that in the future WASI interfaces could be defined in terms of WebIDL (see https://github.com/WebAssembly/WASI/issues/31), in which case we may want to look at auto-generating some of this code from a WebIDL binding generator rather than a proc macro. However for now, this fills a need and looks reasonably straightforward, so let's go with it!

wasi-common-cbindgen/Cargo.toml
kubkon: Excellent point! I've fixed it now to be a dev-dep. Thanks!

wasi-common-cbindgen/src/lib.rs
kubkon: Ah, right, thanks for explaining that :-) Personally, I'm more inclined towards `wasi_common_` than `wasi_` since the crate is called `wasi-common`. Having said that though, it shouldn't be too problematic to change that in the future should we decide to do so.

wasi-common-cbindgen/Cargo.toml
kubkon: Hmm, that's a very good question actually. AFAIK it's just a compile-time dependency. And a quick test reveals that both `.a` and `.rlib` are of similar sizes. `.dylib` is larger now but that's because we're actually exporting the C functions.

kubkon:
> Wow, this is really cool.

Glad you like it! :-)

> It is possible that in the future WASI interfaces could be defined in terms of WebIDL (see [WebAssembly/WASI#31](https://github.com/WebAssembly/WASI/issues/31)), in which case we may want to look at auto-generating some of this code from a WebIDL binding generator rather than a proc macro. However for now, this fills a need and looks reasonably straightforward, so let's go with it!

I've read through the entire conversation and WebIDL looks very appealing to me. In fact, if we could create something like [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen), that would be really cool! So, when it comes to it, I'll be more than happy to help out in those efforts! :-)

wasi-common-cbindgen/src/lib.rs
sunfishcode: `wasi_common_` sounds good to me!

PR #17, opened 2019-05-16
Re-enable cargo cache in travis ci

kubkon:
Reverts 5ff4f57f9 which was added to properly update cargo deps in Travis CI when merging #15. Otherwise, the build would fail at `wasi-common-cbindgen-tests` with a linking error.


sunfishcode:
LGTM

PR #18, opened 2019-05-17
Import all changes from lucet-wasi

kubkon:
I've tried my best to import and merge all changes from lucet-wasi which landed with fastly/lucet#176. There is still some unsafe-ness left that could probably be taken care of (see [here](https://github.com/kubkon/wasi-common/blob/sync-with-lucet/src/hostcalls.rs#L1013)).

I'd be indebted if you could proofread it all! :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 [Wasmtime]: https://github.com/CraneStation/wasmtime
 [Lucet]: https://github.com/fastly/lucet
 [lucet-wasi]: https://github.com/fastly/lucet/tree/master/lucet-wasi
-[lucet-wasi-tracker]: https://github.com/fastly/lucet/commit/5d3efb6005391a7c71d585732a5507b00db6bb1e
+[lucet-wasi-tracker]: https://github.com/fastly/lucet/commit/40ae1df64536250a2b6ab67e7f167d22f4aa7f94
 [WASI API]: https://github.com/CraneStation/wasmtime/blob/master/docs/WASI-api.md
 
 This repo will ultimately serve as a library providing a common implementation of
@@ -16,7 +16,7 @@ WASI hostcalls for re-use in any WASI (and potentially non-WASI) runtimes
 such as [Wasmtime] and [Lucet].
 
 The library is an adaption of [lucet-wasi] crate from the [Lucet] project, and it is
-currently based on [5d3efb6005][lucet-wasi-tracker] git revision.
+currently based on [40ae1df][lucet-wasi-tracker] git revision.
 
 Please note that the library requires Rust compiler version at least 1.34.0.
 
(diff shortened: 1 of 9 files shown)


jedisct1:
Ignoring the formatting issue that Travis chokes on, this looks good to me!

jedisct1 (approved):

README.md
```diff
@@ -25,28 +25,47 @@ Please note that the library requires Rust compiler version at least 1.34.0.
 We support a subset of the [WASI API], though we are working on new hostcalls
 on a regular basis. We currently implement:
```
sunfishcode: If this is all the remaining host calls, can we drop the "subset" here now? Also, can we drop the list of functions here?

src/hostcalls.rs
```diff
@@ -247,167 +240,125 @@ pub fn fd_close(wasi_ctx: &mut WasiCtx, fd: wasm32::__wasi_fd_t) -> wasm32::__wa
...
+        left -= vec_len;
+    }
+    enc_usize_byref(memory, nread, host_nread)
         .map(|_| wasm32::__WASI_ESUCCESS)
         .unwrap_or_else(|e| e)
 }
 
 #[wasi_common_cbindgen]
-pub fn fd_prestat_get(
+pub fn fd_pwrite(
```
sunfishcode: In the Lucet patch, hostcalls.rs was split into hostcalls/fs.rs hostcalls/misc.rs. That seems a desirable direction to head in, both for general code maintenance, and since at the WASI Subgroup level there seems to be broad support for splitting filesystem and other parts of the API into separate modules. Would it be difficult to do a similar split here?

src/hostcalls.rs
```diff
@@ -812,310 +1321,389 @@ pub fn path_filestat_get(
...
+pub fn fd_prestat_get(
     wasi_ctx: &WasiCtx,
     memory: &mut [u8],
     fd: wasm32::__wasi_fd_t,
-    size: wasm32::__wasi_filesize_t,
+    prestat_ptr: wasm32::uintptr_t,
 ) -> wasm32::__wasi_errno_t {
-    unimplemented!("fd_filestat_set_size")
+    let fd = dec_fd(fd);
+    // TODO: is this the correct right for this?
```
sunfishcode: It looks like this TODO comment was removed in the Lucet patch, but I actually think it's something I'd like to think about more, so I'm actually happy you left it in :-).

sunfishcode (commented):
This was a complex merge, with the source path moving a bunch of things around, and the destination patch having some conflicting changes! It looks good overall.

README.md
kubkon: Sure, but we're still missing these:
* `proc_raise`
* `sock_recv`
* `sock_send`
* `sock_shutdown`

sunfishcode: Ah, right. In that case, maybe we can just say we support everything except those, and annotate them:
 - `proc_raise` - we expect this will be removed from WASI, as it isn't very useful without full signal handling support
 - `sock_recv`, `sock_send`, `sock_shutdown` - networking support is not implemented yet

jedisct1: How would you go about implementing `proc_raise` in a generic way? Setting an actual signal handler is not really an option in runtimes that don't run wasm modules in a dedicated process. 

I didn't implement the `sock_*()` hostcalls since there's not much that can be done with them yet. We can't accept() a client, we can't connect(), we can't get any information about a socket. So, even as a stopgap until we come up with a `wasi.net` module, this set of functions doesn't seem to be very useful.

src/hostcalls.rs
kubkon: I've reorganised the hostcalls into submodules like in lucet-wasi. In terms of missing `sock_` hostcalls, I've put them into a submodule of their own `sock.rs`.

README.md
kubkon: As mentioned below, I've put the `sock_` hostcalls into a separate submodule `hostcalls/sock.rs` with `unimplemented!` stubs in them. I reckon this submodule can act as a placeholder until `wasi.net` lands.

In terms of `proc_raise`, if you guys want, I can chuck it out altogether. Lemme know what's your preference on that front! :-)

kubkon: OK, I've now put two things in the README:
1) that we support all hostcalls with the expection of sockets; but we cannot do anything about this until `wasi.net` lands
2) we do not support `proc_raise` as it is expected to be dropped from WASI altogether

jedisct1: Sounds good!

jedisct1 (approved):

sunfishcode:
Looks great!

PR #20, opened 2019-05-19
Fix various failures in misc-tests testsuite

sunfishcode:
This ports a few bug fixes from Wasmtime that hadn't made it into wasi-common yet, and fixes a few additional failures in [misc-tests](https://github.com/sunfishcode/misc-tests), though more failures remain.

Do you think it's worth creating tests for these in the wasi-common repo, or is it sufficient to have them tested by outside testsuites? I'm imagining we can move misc-tests into the Wasmtime repo, to make it easier to run regularly.


kubkon (approved):
LGTM!

kubkon:
As far as tests are concerned, if it wasn't too prohibitive to add some tests to the crate, then I'm all up for it. I think it's going to become even more important as we expand the host support from *nix to other OSes such Windows, etc.

In fact, as I am currently working on reorganising the crate with that in mind (PR coming shortly), and slowly adding some basic functionality for Windows compat, I can really appreciate how the tests would come in handy here. However, if adding tests into the crate is a rather complex and convoluted task, then I'm also fine for having external tests, embedded in Wasmtime (or other runtimes, such as Lucet), etc.

PR #21, opened 2019-05-20
Reorganise to allow for different hosts

kubkon:
This is a rather massive PR for which I apologise, but I couldn't find a better way of dealing with the problem at hand: reorganising to allow for Windows implementation of the crate. Having said that, this PR is by no means final, so if anything is amiss or should be shuffled around, I'll be more than happy to make any necessary changes.

In order to simplify traversing the PR, here's an extended summary of the most important changes:
* wherever possible, I've tried to extract anything common between *nix and Windows in order to avoid duplicate code as much as possible
* the actual host specific implementation is part of a specific submodule of the `sys` module -- this is entirely based on the implementation of Rust's [libstd](https://github.com/rust-lang/rust/tree/master/src/libstd) crate; so, for *nix, its implementation goes in `sys/unix` while for Windows, that would be `sys/windows`
* the host specific implementation is resolved at compile time using [cfg-if](https://github.com/alexcrichton/cfg-if) crate
* `WasiCtx` remains common for all host implementations, however, `FdEntry` is now host specific since, on *nix, we deal with `RawFd` while on Windows with `RawHandle`
* if a hostcall would feature any host specific implementation, I've decided to put it in `sys/{os}` submodule even if there could potentially be some overlap between implementations for different hosts; this one is not set in stone, however, it does make the overall implementation somewhat cleaner
* alas, there does not seem to exist an equivalent of [nix](https://github.com/nix-rust/nix) for Windows; the closest one will get is the [winapi-rs](https://github.com/retep998/winapi-rs) which in terms of functionality and safety is much closer to `libc` than to `nix`
* bearing in mind the previous point, I've ported from Rust nightly the IO vector struct, `IoVec` and `IoVecMut`, which is essentially a wrapper for Winapi's [WSABUF](https://docs.microsoft.com/en-us/windows/desktop/api/ws2def/ns-ws2def-_wsabuf); NB this will become obsolete when [IoSlice/IoSliceMut](https://github.com/rust-lang/rust/pull/60334) land in `libstd` stable

A word of caution here that the implementation of hostcalls on Windows is barely begun, and while it compiles fine, and it's possible to run a simple "Hello world!" WASM binary via [Wasmtime](https://github.com/cranestation/wasmtime/tree/wasi-common), it's still experimental and very much work-in-progress. I'll strive to add more hostcalls and stabilise the implementation over the course of the coming weeks though!

Finally, I've ported `format-all.sh` and `test-all.sh` to Windows (see [format-all.bat](https://github.com/kubkon/wasi-common/blob/nix-win-reorg/format-all.bat) and [test-all.bat](https://github.com/kubkon/wasi-common/blob/nix-win-reorg/test-all.bat)). This made it pretty simple to add Appveyor config for the crate (I've already tested it in my local fork: works as expected).


src/hostcalls/misc.rs
```diff
@@ -77,75 +73,17 @@ pub fn clock_res_get(
...
-            // make sure we follow the spec
-            if resolution == 0 {
-                wasm32::__WASI_EINVAL
-            } else {
-                enc_timestamp_byref(memory, resolution_ptr, resolution)
-                    .map(|_| wasm32::__WASI_ESUCCESS)
-                    .unwrap_or_else(|e| e)
-            }
-        })
+    hostcalls_impl::clock_res_get(memory, clock_id, resolution_ptr)
```
sunfishcode: These wrappers around `hostcalls_impl::` are a fair amount of boilerplate. Would it be possible to have the `clock_res_get` etc. in unix/hostcalls/misc.rs and windows/hostcalls/misc.rs be the actual `clock_res_get` API entry point, instead of having a wrapper here?

sunfishcode (commented):
Overall I like how this looks. Just one concern below:

src/hostcalls/misc.rs
kubkon: Hmm, that's an excellent point. Initially I've had this boilerplate generated with `macro_rules!` macro, but having the actual implementation an API entry point sounds even better. Let me see what I can do about that!

kubkon: OK, I've reshuffled things a little, and now host specific implementation serves also as an API entry point. The common hostcalls reside now in `src/hostcalls.rs`, and also within that module, the host specific hostcalls are publicly re-exported using `pub use crate::sys::hostcalls::*` so that the lib's client code can access all hostcalls from the `wasi_common::hostcalls` module.

src/hostcalls.rs
```diff
@@ -0,0 +1,160 @@
+#![allow(non_camel_case_types)]
+use crate::ctx::WasiCtx;
+use crate::memory::*;
+use crate::wasm32;
+
+use cast::From as _0;
```
sunfishcode, line 7: Please copy the comment [here](https://github.com/japaric/cast.rs/blob/feda734e992e491e03c184b3d5d21e00917c4a9c/src/lib.rs#L53) here so that it's clear what this is for.

src/sys/unix/hostcalls/fs.rs
```diff
@@ -1,8 +1,10 @@
 #![allow(non_camel_case_types)]
 #![allow(unused_unsafe)]
+use crate::sys::fdentry::{determine_type_rights, FdEntry};
+use crate::sys::host as host_impl;
```
sunfishcode: Would it make sense to rename sys/unix/host.rs and sys/windows/host.rs to .../host_impl.rs, so that we don't have to use an "as" here?

src/sys/unix/host.rs
```diff
@@ -0,0 +1,240 @@
+//! WASI host types specific to *nix host.
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+#![allow(dead_code)]
+use crate::host;
+
+pub fn errno_from_nix(errno: nix::errno::Errno) -> host::__wasi_errno_t {
```
sunfishcode: It seems `errno_from_nix` changed from returning a `wasm32::__wasi_errno_t` to a `host::__wasi_errno_t`. However, I think the direction we're heading is that for most types, this distinction doesn't matter, and we should pull them out of "host" and "wasm32" anyway -- though that's a separate refactoring we can do. So I think we can leave this code as-is for now.

src/hostcalls.rs
kubkon: Thanks! It's done now.

src/sys/unix/hostcalls/fs.rs
kubkon: It certainly would, thanks! I've done it now :-)

src/sys/unix/host.rs
kubkon: Yeah, this is the one I've had some problems in categorising. As you pointed out, on one hand, it was returning `wasm32::__wasi_errno_t`, but on the other, it takes `nix::errno::Errno` as an argument which to me belongs in host specific implementation. Hence why I decided to unify the two versions of the `errno_from_nix` functions. Also, as you rightly mentioned here, and as far as I remember also in fastly/lucet#176, we seem to be heading in the direction of unifying `host` and `wasm32` anyway, so it seemed like a prudent thing to do here ;-)

kubkon:
Just some heads up about formatting and `rustfmt` in general. It turns out that there is a bug in `rustfmt` (rust-lang/rustfmt#3253) which prevents `rustfmt` from formatting modules inside a macro such as [cfg-ig](https://github.com/alexcrichton/cfg-if).

This seems pretty serious and difficult to solve, and I'm not sure how we should deal with it until a potential fix is found. For this PR, for example, I've commented out each conditional route one at a time in `src/sys/mod.rs` and then re-run `cargo fmt`. This works but obviously is less than ideal. The interesting thing is that even [libc](https://github.com/rust-lang/libc) seems affected, and I'd imagine that also Rust's `libstd` was affected as well.

sunfishcode:
Cool, this looks in good shape!

PR #23, opened 2019-05-31
Backport path_get fix discovered in Win branch

kubkon:
Otherwise, for nested directories, we were trying an `openat` wrt the preopened fd always (regardless how far we've advanced down the path) rather than from the most recently opened fd. As a result, for a nested dir path like:

```
{preopened_fd}/a/b/c
```

it would result in a `ENOENT` error when processing `b/c`, since we'd try to open `b` also wrt `{preopened_fd}`:

```
{preopened_fd}/b/c
```

rather than wrt to freshly opened `a`:

```
a/b/c
```

### DIFF
--- a/src/sys/unix/hostcalls/fs_helpers.rs
+++ b/src/sys/unix/hostcalls/fs_helpers.rs
@@ -149,7 +149,7 @@ pub fn path_get<P: AsRef<OsStr>>(
                     || (component.ends_with(b"/") && !needs_final_component) =>
             {
                 match openat(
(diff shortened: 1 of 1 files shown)


jedisct1 (approved):
Good catch!

kubkon:
@jedisct1 lucky catch actually. Surfaced when I was implementing `path_get` on Windows. I actually really like the idea of having fd and path stacks and advancing down the actual path using `openat` pushing and popping the stacks here and there - simply brilliant! :-)

PR #24, opened 2019-05-31
Use `try_from` instead of the `cast` crate.

sunfishcode:
`try_from` is newly stable in the standard library starting in Rust 1.34.


kubkon (approved):
LGTM!

PR #26, opened 2019-06-11
Move common functionality into hostcalls mod

kubkon:
This PR is essentially a housekeeping change. It strives to refactor out as much of common functionality between *nix and other hosts such as Windows as possible.

To provide some context, this PR came about as a direct result of my discussion with @sunfishcode (see #22) that it would be nice to re-use as much of common stuff as possible especially since WASI is still rather volatile and certain things which might be common between different hosts might change in the near future.

Quick summary of changes:
* common functionality has now been moved to a separate module `hostcalls` which is located directly in the root of the crate:
* each hostcall exposed by the lib to the outside world now resides in this separate module, and internally, most of the hostcalls call appropriately *nix or Windows specific impls


sunfishcode:
This looks great! It is likely that as WASI grows, we'll have even more code in common between the platforms, so this is a good direction to head in.

kubkon:
> This looks great! It is likely that as WASI grows, we'll have even more code in common between the platforms, so this is a good direction to head in.

Agreed! :-)

PR #27, opened 2019-06-11
Nix fixes

kubkon:
This PR builds upon #26. It patches up all problems that cause [sunfishcode/misc-tests](https://github.com/sunfishcode/misc-tests) to fail.

Summary of fixes:
* `fd_renumber` should currently fail when trying to renumber a preopen
* `path_readlink` should be able to handle 0-sized buffer (as pointed out by @sunfishcode [in the original C impl](https://github.com/CraneStation/wasmtime/blob/master/wasmtime-wasi-c/sandboxed-system-primitives/src/posix.c#L1955), Linux requires the input buffer to have positive size, however, POSIX does not enforce that)
* `path_open` had the `OFlag`s the other way around: the `else` branch when determining read/write attributes should have `OFlags::O_RDONLY` as a fallthrough and not `OFlags::O_WRONLY` as it was until now
* `path_get` now correctly handles "interesting paths" with a plethora of trailing slashes, "..", etc. To achieve this, instead of iterating over raw paths (using bytes), we use `std::path::Components` functionality instead; it is still necessary to preserver the path's trailing slash so that opening a file as a dir fails as expected
* `path_get` now correctly fails with `EILSEQ` in the presence of spurious NULs in the input path
* applies fix CraneStation/wasmtime#173 to `path_open`
* fixes a typo in `path_open` to do with setting the `needed_base` flags in the presence of `OFlag::O_TRUNC`: up to now we were incorrectly adjusting `needed_inheriting` flags instead of `needed_base`
* finally, `path_unlink_file` now correctly handles non-Linux *nix OSes, i.e., `unlinkat` may return `EPERM` on non-Linux host and it should be adjusted to `EISDIR`

### DIFF
--- a/src/sys/unix/hostcalls_impl/fs.rs
+++ b/src/sys/unix/hostcalls_impl/fs.rs
@@ -516,18 +516,43 @@ pub(crate) fn path_readlink(
     rights: host::__wasi_rights_t,
     buf: &mut [u8],
 ) -> Result<usize, host::__wasi_errno_t> {
-    use nix::fcntl::readlinkat;
+    use nix::errno::Errno;
 
     let (dir, path) = match path_get(wasi_ctx, dirfd, 0, path, rights, 0, false) {
         Ok((dir, path)) => (dir, path),
         Err(e) => return Err(e),
     };
 
-    let target_path = match readlinkat(dir, path.as_os_str(), buf) {
-        Err(e) => return Err(host_impl::errno_from_nix(e.as_errno().unwrap())),
-        Ok(target_path) => target_path,
+    let path_cstr = match std::ffi::CString::new(path.as_bytes()) {
+        Ok(path_cstr) => path_cstr,
+        Err(_) => return Err(host::__WASI_EINVAL),
     };
-    Ok(target_path.len())
+
+    // Linux requires that the buffer size is positive, whereas POSIX does not.
+    // Use a fake buffer to store the results if the size is zero.
+    // TODO: instead of using raw libc::readlinkat call here, this should really
+    // be fixed in `nix` crate
+    let fakebuf: &mut [u8] = &mut [0];
+    let buf_len = buf.len();
+    let len = unsafe {
+        libc::readlinkat(
+            dir,
+            path_cstr.as_ptr() as *const libc::c_char,
+            if buf_len == 0 {
+                fakebuf.as_mut_ptr()
+            } else {
+                buf.as_mut_ptr()
+            } as *mut libc::c_char,
+            if buf_len == 0 { fakebuf.len() } else { buf_len },
+        )
+    };
+
+    if len < 0 {
+        Err(host_impl::errno_from_nix(Errno::last()))
+    } else {
+        let len = len as usize;
+        Ok(if len < buf_len { len } else { buf_len })
+    }
 }
 
 pub(crate) fn path_rename(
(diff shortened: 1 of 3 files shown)


jedisct1 (approved):

src/sys/unix/hostcalls_impl/fs.rs
```diff
@@ -0,0 +1,867 @@
...
+            dir,
+            path_cstr.as_ptr() as *const libc::c_char,
+            if buf_len == 0 {
+                fakebuf.as_mut_ptr()
+            } else {
+                buf.as_mut_ptr()
+            } as *mut libc::c_char,
+            if buf_len == 0 { fakebuf.len() } else { buf_len },
+        )
+    };
```
sunfishcode, line 548: The fix for the zero-length readlink buffer in https://github.com/fastly/lucet/pull/213 looks a little simpler; would it make sense to do the same thing here? That said, I do like the comment here about investigating having `nix` take care of this detail transparently, so please preserve that comment in any case :-).

sunfishcode (commented):
Very cool!

src/sys/unix/hostcalls_impl/fs.rs
kubkon: Hmm, do you know if they've double checked it for this particular use case? I'm asking because actually, I've already tried something similar before, but due to the way `nix::fcntl::readlinkat` is implemented, it wouldn't work for me. Perhaps I was doing something wrong. But to elaborate, here's the `readlinkat` impl taken verbatim from the `nix` crate (here's the [link](https://github.com/nix-rust/nix/blob/master/src/fcntl.rs#L180)):

```rust
pub fn readlinkat<'a, P: ?Sized + NixPath>(dirfd: RawFd, path: &P, buffer: &'a mut [u8]) -> Result<&'a OsStr> {
    let res = path.with_nix_path(|cstr| {
        unsafe { libc::readlinkat(dirfd, cstr.as_ptr(), buffer.as_mut_ptr() as *mut c_char, buffer.len() as size_t) }
    })?;

    wrap_readlink_result(buffer, res)
}
```

And in turn `wrap_readlink_result` does the following:

```rust
fn wrap_readlink_result(buffer: &mut[u8], res: ssize_t) -> Result<&OsStr> {
    match Errno::result(res) {
        Err(err) => Err(err),
        Ok(len) => {
            if (len as usize) >= buffer.len() {
                Err(Error::Sys(Errno::ENAMETOOLONG))
            } else {
                Ok(OsStr::from_bytes(&buffer[..(len as usize)]))
            }
        }
    }
}
```

Due to the `wrap_readlink_result`, if we ever try to read link that's longer than the buffer we provide, `nix` will trip with `ENAMETOOLONG` which is exactly the error I was getting when trying the same approach as fastly/lucet#213 for zero-length buffer.

@jedisct1, @sunfishcode any thoughts and help on this will be much appreciated! :-)

kubkon: In the comment above, when I said "it wouldn't work for me" I meant that fastly/lucet#213 approach wouldn't pass the `test_readlink_no_buffer` test in [sunfishcode/misc-tests](https://github.com/sunfishcode/misc-tests/blob/master/src/main.rs#L599). As I pointed out above, this is due to the fact that, the `nix` crate doesn't only wrap `libc::readlinkat` syscall but actually checks whether the passed in buffer is larger than the actual number of bytes written to the buffer. So, even if we pass in a buffer of length 1, due to the above additional check, I believe that `nix::fcntl::readlinkat` will always return `ENAMETOOLONG` for that particular case.

sunfishcode:
Ok, the PR here looks good to me. I'm going to file a separate issue about `readlinkat` so we can continue to track that without holding up other progress.

sunfishcode:
I've now filed https://github.com/CraneStation/wasi-common/issues/30.

kubkon:
> I've now filed #30.

Nice one, thanks!

PR #31, opened 2019-06-27
Use Travis CI to build on all OSes

kubkon:
This is a quick change and cleanup to see if Travis can do it for all 3 major players: linux, osx, and win. A free bonus of this change is that we no longer need separate batch scripts for win; we can simply re-use the shell scripts instead.

Anyhow, this is also an experiment to check how Travis fares against Azure Pipelines which is getting set up in [CraneStation/wasi-libc](https://github.com/CraneStation/wasi-libc) (see CraneStation/wasi-libc#78).


jedisct1:
This is great!

Why are we explicitly installing rust `1.34.0` instead of `stable` btw?

kubkon:
> This is great!
> 
> Why are we explicitly installing rust `1.34.0` instead of `stable` btw?

Good catch! I don't see any reason why we couldn't have `stable` instead of explicitly specified version as is now. I'll change it to `stable` then :-)

jedisct1 (approved):

.travis.yml
```diff
@@ -1,43 +1,42 @@
...
-    - osx
+  - linux
+  - osx
+  - windows
 language: rust
 rust:
-    - 1.34.0
-    - beta
-    - nightly
+  - stable
```
sunfishcode: As a library meant to be used by other projects, it can be useful to document (and test) the minimum supported Rust version. This was explicitly `1.34` because there's a badge in the top-level README which advertises Rust 1.34 support.

I recognize this likely isn't very important for wasi-common right now, but this practice comes from Cranelift, which does have users which sometimes care about releases older than `stable`.

.travis.yml
```diff
@@ -1,43 +1,42 @@
...
+  - rustup component list --toolchain=stable
+  - rustup show
+  - rustfmt +stable --version || echo fail
+  # Sometimes the component isn't actually ready after being installed, and
+  # rustup update makes it ready.
+  - rustup update
+  - rustfmt +stable --version
 script: ./test-all.sh
-cache:
-    cargo: true
```
sunfishcode, line 43: Why is `cargo` caching removed?

sunfishcode (commented):
I don't know much about Travis CI's Windows support vs Appveyer vs Azure Pipelines, but I do know it's nice to not have two copies of the format-all/test-all scripts :-).

.travis.yml
kubkon: Ah, of course! In that case, let me add `1.34` back but at the same time preserve `stable` as well. This way, we'll test against both the minimum supported as well as the latest stable.

.travis.yml
kubkon: Oh crap, I was testing some things on travis without cache and forgot to revert the topmost commit to re-enable it. Thanks for spotting it!

kubkon: Just FYI, I've decided to use a more specialised caching strategy than the default provided by Travis. Namely, here's the relevant bit of `.travis.yml`:

```
cache:
  directories:
    - $HOME/.cargo
before_cache:
  - rm -rf /home/travis/.cargo/registry
```

This strategy is apparently meant to offer the best performance according to this fantastic [blog post](https://levans.fr/rust_travis_cache.html). I've also noticed that it seems to prevent weird linking errors in Travis while testing `wasi-common-cbindgen` so that's also a plus.

kubkon:
@tschneidereit @sunfishcode As discussed offline, I'll merge these change in, since the setup work has already been done. However, in the meantime, to follow Rust's suit, I'll start figuring out how to migrate to Azure. @tschneidereit that is of course not unless you've already got something setup, then I'll be happy to help out! :-)

PR #32, opened 2019-06-28
Fix inheriting stdio on Windows

kubkon:
This PR fixes failures on CIs with the recent changes to Windows implementation (you can see an example of such a failure [here](https://travis-ci.org/CraneStation/wasi-common/jobs/551644215)). As it turns out, the CI in Windows builds pipes the `stdio` handles raising an `ERROR_INVALID_HANDLE` error when trying to duplicate them on Windows.

The proposed fix to this problem is to create a handle to `stdio` manually (using `CreateFile` winapi call rather than relying on `GetStdHandle` as it's customary; in Rust terms, we'd use `OpenOptions` to create a new handle to `STDIN` and `STDOUT` rather than relying on `Stdin` and `Stdout` structs). This is in accordance with the official [MSDN docs](https://docs.microsoft.com/en-us/windows/console/console-handles):

> the SetStdHandle function can redirect the standard handles by changing the handle associated with STDIN, STDOUT, or STDERR. Because the parent's standard handles are inherited by any child process, subsequent calls to GetStdHandle return the *redirected* handle.

> The CreateFile function enables a process to get a handle to its console's input buffer and active screen buffer, even if STDIN and STDOUT have been redirected. To open a handle to a console's input buffer, specify the CONIN$ value in a call to CreateFile. Specify the CONOUT$ value in a call to CreateFile to open a handle to a console's active screen buffer.

However, I welcome some discussion and suggestions here on what the best approach to this problem would be.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ cranelift-entity = "0.30.0"
 cranelift-wasm = "0.30.0"
 cranelift-native = "0.30.0"
 target-lexicon = "0.3.0"
+pretty_env_logger = "0.3.0"
 
 [patch."https://github.com/CraneStation/wasi-common"]
 wasi-common = { path = "." }
--- a/build.rs
+++ b/build.rs
@@ -55,7 +55,7 @@ fn test_directory(out: &mut File, testsuite: &str) -> io::Result<()> {
             .expect("testsuite filename should be representable as a string")
             .replace("-", "_")
     )?;
-    writeln!(out, "    use super::{{runtime, utils}};")?;
+    writeln!(out, "    use super::{{runtime, utils, setup_log}};")?;
     for dir_entry in dir_entries {
         write_testsuite_tests(out, dir_entry, testsuite)?;
     }
@@ -80,6 +80,7 @@ fn write_testsuite_tests(out: &mut File, dir_entry: DirEntry, testsuite: &str) -
         "    fn {}() -> Result<(), String> {{",
         avoid_keywords(&stemstr.replace("-", "_"))
     )?;
+    write!(out, "        setup_log();")?;
     write!(out, "        let path = std::path::Path::new(\"")?;
     // Write out the string with escape_debug to prevent special characters such
     // as backslash from being reinterpreted.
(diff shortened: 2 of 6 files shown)


kubkon:
cc @alexcrichton

alexcrichton:
This doesn't feel quite right to me since `SetStdHandle` can be used to override stdin/stdout for an application but opening the files I think opens the stdin/stdout for the terminal, so it may cause surprising behavior when output is intended to be redirected.

If the issue is just `duplicate`, does this even need to use `DuplicateHandle` or `dup`? Could it simply imply reference counting on the host?

kubkon:
> This doesn't feel quite right to me since `SetStdHandle` can be used to override stdin/stdout for an application but opening the files I think opens the stdin/stdout for the terminal, so it may cause surprising behavior when output is intended to be redirected.
> 

That's correct. We do open stdin/stdout for Windows Console device here if `SetStdHandle` was used to redirect the stream. I thought that the proposed solution would break the simple print out to the "screen" in Travis/Appveyor after this fix, but surprisingly, this is not the case, and everything gets logged in the CI just fine.

> If the issue is just `duplicate`, does this even need to use `DuplicateHandle` or `dup`? Could it simply imply reference counting on the host?

Yep, my thinking exactly. Just to confirm that we're on the same page here: rather than duplicating the handle/fd to stdio, you'd rather simply store a reference to the "stdio" handles/fds? If that's the case, to me it also seems like a prefered approach.

kubkon:
@alexcrichton I was thinking, perhaps we could wrap `std::io::Stdin` (and others) algebraically in a way similar to:

```rust
pub(crate) enum Descriptor {
    File(ManuallyDrop::<fs::File>),
    Stdin,
    Stdout,
    Stderr,
}

pub(crate) struct FdObject {
    file_type: host::__wasi_filetype_t,
    descriptor: Descriptor,
    needs_close: bool,
}
```

Perhaps this would solve the duplication issue here. What do you reckon?

PS. The above suggestion is part of a rewrite which reuses as much `libstd` as possible which you can inspect [here](https://github.com/CraneStation/wasi-common/compare/master...kubkon:alex-rewrite). Bear in mind, it's very incomplete and will most likely not compile yet!

kubkon:
OK, I’m going to close this PR in favour of #34 which also fixes the issue but in a less hacky way. In fact, it strives to follow @alexcrichton’s suggestion to count references to stdio handles rather than duplicating them on the host by reusing `libstd` structs.

PR #33, opened 2019-06-29
Use ManuallyDrop instead of forget

kubkon:
Backports a fix suggested by @alexcrichton in #22  to *nix impl.


sunfishcode:
Nice!

PR #38, opened 2019-07-19
Further refactor of hostcalls

kubkon:
This PR does further refactoring of the hostcalls. In particular, it
* introduces helper type `Result<T> = Result<T, host::__wasi_errno_t>`
* swaps `&FdEntry` for `&File` wherever possible

As a by-product of the second bulletpoint, `fd_renumber` is now platform independent. I've also added a positive testcase for `fd_renumber` to verify that it indeed is correct.

EDIT: ~~**NB** while this PR builds on #37, it is not a prerequisite for it, and should we decide to go with pure `&str` and `String` instead of `WasiPath` and `WasiPathBuf` wrapper types as per our discussion in #37, I'll make all the necessary rebasing work, so don't let this sway you when going over this PR :-)~~


kubkon:
cc @sunfishcode @alexcrichton

alexcrichton:
This seems reasonable to me! I ended up seeing #39 first and left one tangential comment there which is probably also equally applicable to this, but both seem fine to land now.

sunfishcode:
Looks good!

PR #39, opened 2019-07-21
Refactor fd_seek and fd_tell

kubkon:
This PR reuses Rust's `libstd` to refactor `fd_seek` and `fd_tell`. In particular, it uses [Seek](https://doc.rust-lang.org/std/io/trait.Seek.html) trait to provide platform-independent implementation.

**NB** since this PR builds on #38, it is a prerequisite for merging.


kubkon:
cc @alexcrichton @sunfishcode

src/ctx.rs
```diff
@@ -147,20 +144,27 @@ impl WasiCtx {
...
+    pub fn new<S: AsRef<str>>(args: impl Iterator<Item = S>) -> Result<Self> {
         WasiCtxBuilder::new()
             .and_then(|ctx| ctx.args(args))
             .and_then(|ctx| ctx.inherit_stdio())
             .and_then(|ctx| ctx.inherit_env())
             .and_then(|ctx| ctx.build())
     }
 
+    pub fn contains_fd_entry(&self, fd: host::__wasi_fd_t) -> bool {
+        match self.fds.get(&fd) {
```
alexcrichton: This could also be `self.fds.contains_key(&fd)`

src/hostcalls/fs.rs
```diff
@@ -278,12 +299,25 @@ pub fn fd_seek(
...
         host::__WASI_RIGHT_FD_SEEK | host::__WASI_RIGHT_FD_TELL
     };
-    let fe = match wasi_ctx.get_fd_entry(fd, rights, 0) {
-        Ok(fe) => fe,
+    let mut fd = match wasi_ctx
+        .get_fd_entry(fd, rights, 0)
+        .and_then(|fe| fe.fd_object.descriptor.as_file())
+    {
+        Ok(f) => f,
         Err(e) => return return_enc_errno(e),
```
alexcrichton, line 307: This is sort of a general style about the implementation of these functions and not particularly relevant to this PR itself, but I think that syscalls probably want to be defined "the majority" in functions returning `Result` for the usage of `?`. It looks like a lot of code is now in functions which has a bunch of `match` statements like this so it may be worthwhile at some point to refactor.

src/ctx.rs
kubkon: Good catch, thanks! I've addressed this now in #38 :-)

src/hostcalls/fs.rs
kubkon: Yeah, that's an excellent point, and already on my todo list (which BTW is mysteriously growing larger the more is done... go figure! 🤔 ). I'll address it in one of the upcoming PRs, but thanks for pointing out!

kubkon: OK, it's done. Have a look at #40 :-)

sunfishcode:
This looks good to me; it just needs some conflicts resolved now.

kubkon:
> This looks good to me; it just needs some conflicts resolved now.

On it :+1:

kubkon:
@sunfishcode done! :-)

sunfishcode:
Thanks!

PR #40, opened 2019-07-22
Move hostcall impl behind a Result<()> fn

kubkon:
This PR addresses @alexcrichton's great suggestion of moving the implementation of the majority of hostcalls into functions which return a `Result<(), host::__wasi_errno_t>`, and thus making use of the `?` operator to streamline a lot of the processing.

**NB** This PR depends directly on #39, and in turn on #38.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,0 +1,13 @@
+macro_rules! hostcalls {
+    ($(pub fn $name:ident($($arg:ident: $ty:ty,)*) -> $ret:ty;)*) => ($(
+            #[wasi_common_cbindgen::wasi_common_cbindgen]
+            pub fn $name($($arg: $ty,)*) -> $ret {
+                let ret = match crate::hostcalls_impl::$name($($arg,)*) {
+                    Ok(()) => crate::host::__WASI_ESUCCESS,
+                    Err(e) => e,
+                };
+
+                crate::hostcalls::return_enc_errno(ret)
+            }
+    )*)
+}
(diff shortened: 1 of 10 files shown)


kubkon:
cc @alexcrichton @sunfishcode

src/macros.rs
```diff
@@ -0,0 +1,13 @@
+macro_rules! hostcalls {
+    ($(pub fn $name:ident($($arg:ident: $ty:ty,)*) -> $ret:ty;)*) => ($(
+            #[wasi_common_cbindgen::wasi_common_cbindgen]
+            pub fn $name($($arg: $ty,)*) -> $ret {
+                let ret = match crate::hostcalls_impl::$name($($arg,)*) {
+                    Ok(_) => crate::host::__WASI_ESUCCESS,
```
alexcrichton: Should this perhaps require that `Ok(())` is matched? That'd help avoid issues where something is accidentally dropped here when it was supposed to be returned.

Ideally we could automatically detect `()` vs-not and have an automatically synthesized outptr, but that's a bit pie-in-the-sky right now

alexcrichton:
Looks reasonable to me! I don't have tons of opinions on organization, but I'm assuming that 99% of the movement here isn't changing code and is just `match` -> `?`, so I've just got one thought about the macro before merging

kubkon:
> Looks reasonable to me! I don't have tons of opinions on organization, but I'm assuming that 99% of the movement here isn't changing code and is just `match` -> `?`, so I've just got one thought about the macro before merging

Yep, you got it!

src/macros.rs
kubkon: Good catch, thanks!

Yep, that'd indeed be amazing!

alexcrichton (approved):

sunfishcode:
Nice!

PR #42, opened 2019-07-24
Implement fd_filestat_get for all platforms

marmistrz:
Current issues/doubts about this PR:
* On Windows, we make multiple syscalls for every call of `fd_filestat_get`. Anyway, so as to implement this function efficiently, we need to move a large part of the functionality to libstd, so I prefer code cleanliness over performance for now in this change
* I'm not sure what to do with integer overflows/failed conversions between various types (currently I'm using `try_into` and panic if the conversion fails)
* `file_allocate` fails on Windows for no clear reason: http://paste.debian.net/1092902/ (this paste will expire in 90d) and it seems not to be connected with my change - it looks like the tests crashes during https://github.com/CraneStation/wasi-misc-tests/blob/9f8c648fb14afc1ed9d2effa2567870d65cbc902/src/bin/file_allocate.rs#L91

### DIFF
--- a/src/hostcalls_impl/fs.rs
+++ b/src/hostcalls_impl/fs.rs
@@ -4,7 +4,7 @@ use crate::fdentry::{Descriptor, FdEntry};
 use crate::memory::*;
 use crate::sys::{errno_from_host, host_impl, hostcalls_impl};
 use crate::{host, wasm32, Result};
-use log::trace;
+use log::{debug, trace};
 use std::io::{self, Read, Seek, SeekFrom, Write};
 
 pub(crate) fn fd_close(wasi_ctx: &mut WasiCtx, fd: wasm32::__wasi_fd_t) -> Result<()> {
@@ -730,11 +730,44 @@ pub(crate) fn fd_filestat_get(
         .get_fd_entry(fd, 0, 0)
         .and_then(|fe| fe.fd_object.descriptor.as_file())?;
 
-    let host_filestat = hostcalls_impl::fd_filestat_get(fd)?;
+    let host_filestat = fd_filestat_get_impl(fd).map_err(|e| {
+        debug!("fd_filestat_get: error: {}", e);
+        e.raw_os_error().map_or(host::__WASI_EIO, errno_from_host)
+    })?;
 
     trace!("     | *filestat_ptr={:?}", host_filestat);
 
     enc_filestat_byref(memory, filestat_ptr, host_filestat)
+}
+
+fn fd_filestat_get_impl(file: &std::fs::File) -> io::Result<host::__wasi_filestat_t> {
+    use std::convert::TryInto;
+    use std::time::{SystemTime, UNIX_EPOCH};
+    fn timestamp(st: SystemTime) -> io::Result<u64> {
+        st.duration_since(UNIX_EPOCH)
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, e.to_string()))?
+            .as_nanos()
+            .try_into()
+            .map_err(|e: std::num::TryFromIntError| {
+                io::Error::new(io::ErrorKind::Other, e.to_string())
+            })
+    }
+    let metadata = file.metadata()?;
+    // On Windows all the information needed is either in libstd or provided by
+    // GetFileInformationByHandle winapi call. All of it is much easier to implement
+    // in libstd
+    Ok(host::__wasi_filestat_t {
+        st_dev: hostcalls_impl::device_id(file)?,
+        st_ino: hostcalls_impl::file_serial_no(file)?,
+        st_nlink: hostcalls_impl::num_hardlinks(file)?
+            .try_into()
+            .expect("overflow"),
+        st_size: metadata.len(),
+        st_atim: metadata.accessed().and_then(timestamp)?,
+        st_ctim: hostcalls_impl::change_time(file)?,
+        st_mtim: metadata.modified().and_then(timestamp)?,
+        st_filetype: hostcalls_impl::filetype(file)?,
+    })
 }
 
 pub(crate) fn fd_filestat_set_times(
--- a/src/hostcalls_impl/fs.rs
+++ b/src/hostcalls_impl/fs.rs
@@ -4,7 +4,7 @@ use crate::fdentry::{Descriptor, FdEntry};
 use crate::memory::*;
 use crate::sys::{errno_from_host, host_impl, hostcalls_impl};
 use crate::{host, wasm32, Result};
-use log::{debug, trace};
+use log::trace;
 use std::io::{self, Read, Seek, SeekFrom, Write};
 
 pub(crate) fn fd_close(wasi_ctx: &mut WasiCtx, fd: wasm32::__wasi_fd_t) -> Result<()> {
@@ -730,53 +730,11 @@ pub(crate) fn fd_filestat_get(
         .get_fd_entry(fd, 0, 0)
         .and_then(|fe| fe.fd_object.descriptor.as_file())?;
 
-    let host_filestat = fd_filestat_get_impl(fd)?;
+    let host_filestat = hostcalls_impl::fd_filestat_get_impl(fd)?;
 
     trace!("     | *filestat_ptr={:?}", host_filestat);
 
     enc_filestat_byref(memory, filestat_ptr, host_filestat)
-}
-
-fn fd_filestat_get_impl(file: &std::fs::File) -> Result<host::__wasi_filestat_t> {
-    use std::convert::TryInto;
-    use std::time::{SystemTime, UNIX_EPOCH};
-
-    fn convert_err(e: io::Error) -> host::__wasi_errno_t {
-        debug!("fd_filestat_get: os error: {}", e);
-        e.raw_os_error().map_or(host::__WASI_EIO, errno_from_host)
-    }
-
-    fn timestamp(st: SystemTime) -> Result<u64> {
-        st.duration_since(UNIX_EPOCH)
-            .map_err(|_| host::__WASI_EINVAL)? // date earlier than UNIX_EPOCH
-            .as_nanos()
-            .try_into()
-            .map_err(|_| host::__WASI_EOVERFLOW) // u128 doesn't fit into u64
-    }
-
-    let metadata = file.metadata().map_err(convert_err)?;
-    Ok(host::__wasi_filestat_t {
-        st_dev: hostcalls_impl::device_id(file, &metadata).map_err(convert_err)?,
-        st_ino: hostcalls_impl::file_serial_no(file, &metadata).map_err(convert_err)?,
-        st_nlink: hostcalls_impl::num_hardlinks(file, &metadata)
-            .map_err(convert_err)?
-            .try_into()
-            .map_err(|_| host::__WASI_EOVERFLOW)?, // u64 doesn't fit into u32
-        st_size: metadata.len(),
-        st_atim: metadata
-            .accessed()
-            .map_err(convert_err)
-            .and_then(timestamp)?,
-        st_ctim: hostcalls_impl::change_time(file, &metadata)
-            .map_err(convert_err)?
-            .try_into()
-            .map_err(|_| host::__WASI_EOVERFLOW)?, // i64 doesn't fit into u64
-        st_mtim: metadata
-            .modified()
-            .map_err(convert_err)
-            .and_then(timestamp)?,
-        st_filetype: hostcalls_impl::filetype(&metadata).map_err(convert_err)?,
-    })
 }
 
 pub(crate) fn fd_filestat_set_times(
--- a/src/hostcalls_impl/fs.rs
+++ b/src/hostcalls_impl/fs.rs
@@ -753,9 +753,6 @@ fn fd_filestat_get_impl(file: &std::fs::File) -> io::Result<host::__wasi_filesta
             })
     }
     let metadata = file.metadata()?;
-    // On Windows all the information needed is either in libstd or provided by
-    // GetFileInformationByHandle winapi call. All of it is much easier to implement
-    // in libstd
     Ok(host::__wasi_filestat_t {
         st_dev: hostcalls_impl::device_id(file)?,
         st_ino: hostcalls_impl::file_serial_no(file)?,
--- a/src/hostcalls_impl/fs.rs
+++ b/src/hostcalls_impl/fs.rs
@@ -761,7 +761,7 @@ fn fd_filestat_get_impl(file: &std::fs::File) -> io::Result<host::__wasi_filesta
             .expect("overflow"),
         st_size: metadata.len(),
         st_atim: metadata.accessed().and_then(timestamp)?,
-        st_ctim: hostcalls_impl::change_time(file)?,
+        st_ctim: hostcalls_impl::change_time(file)?.try_into().expect("overflow"),
         st_mtim: metadata.modified().and_then(timestamp)?,
         st_filetype: hostcalls_impl::filetype(file)?,
     })
--- a/src/hostcalls_impl/fs.rs
+++ b/src/hostcalls_impl/fs.rs
@@ -756,9 +756,9 @@ fn fd_filestat_get_impl(file: &std::fs::File) -> Result<host::__wasi_filestat_t>
 
     let metadata = file.metadata().map_err(convert_err)?;
     Ok(host::__wasi_filestat_t {
-        st_dev: hostcalls_impl::device_id(file).map_err(convert_err)?,
-        st_ino: hostcalls_impl::file_serial_no(file).map_err(convert_err)?,
-        st_nlink: hostcalls_impl::num_hardlinks(file)
+        st_dev: hostcalls_impl::device_id(file, &metadata).map_err(convert_err)?,
+        st_ino: hostcalls_impl::file_serial_no(file, &metadata).map_err(convert_err)?,
+        st_nlink: hostcalls_impl::num_hardlinks(file, &metadata)
             .map_err(convert_err)?
             .try_into()
             .map_err(|_| host::__WASI_EOVERFLOW)?, // u64 doesn't fit into u32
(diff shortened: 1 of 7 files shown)


kubkon:
Thanks! I've not had a chance to look at it in-detail just yet, but a couple of high-level comments:
* we've moved away from panics entirely, and would prefer to return an appropriate `host::__wasi_errno_t` instead; for instance, for overflows I'd suggest to throw a `host::__WASI_EOVERFLOW` error while for failed conversions `host::__WASI_EILSEQ` or `host::__WASI_EINVAL`
* as far as the failed testcases are concerned, you can analyse the trace of syscalls in-detail with the following command:

```
RUST_LOG=wasi_common=trace cargo test file_allocate
```

In particular, for `file_allocate`, it is expected to fail on Windows until #41 and in particular `path_unlink_file` is implemented :-)

build.rs
```diff
@@ -136,6 +136,7 @@ cfg_if::cfg_if! {
                     "file_pread_pwrite" => false,
                     "renumber" => false,
                     "file_seek_tell" => false,
+                    "file_allocate" => false,
```
kubkon: This will not run on Windows successfully just yet so please revert this change for now :-)

src/hostcalls_impl/fs.rs
```diff
@@ -730,13 +730,46 @@ pub(crate) fn fd_filestat_get(
...
 
     enc_filestat_byref(memory, filestat_ptr, host_filestat)
 }
 
+fn fd_filestat_get_impl(file: &std::fs::File) -> io::Result<host::__wasi_filestat_t> {
+    use std::convert::TryInto;
+    use std::time::{SystemTime, UNIX_EPOCH};
+    fn timestamp(st: SystemTime) -> io::Result<u64> {
+        st.duration_since(UNIX_EPOCH)
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, e.to_string()))?
```
kubkon: Instead of mapping into catchall `io::ErrorKind::Other` here, wouldn't it be better to map directly to `host::__WASI_EINVAL` for example?

src/hostcalls_impl/fs.rs
```diff
@@ -730,13 +730,46 @@ pub(crate) fn fd_filestat_get(
...
+            .as_nanos()
+            .try_into()
+            .map_err(|e: std::num::TryFromIntError| {
+                io::Error::new(io::ErrorKind::Other, e.to_string())
```
kubkon: Same here, as above :-)

src/hostcalls_impl/fs.rs
```diff
@@ -730,13 +730,46 @@ pub(crate) fn fd_filestat_get(
...
+    let metadata = file.metadata()?;
+    // On Windows all the information needed is either in libstd or provided by
+    // GetFileInformationByHandle winapi call. All of it is much easier to implement
+    // in libstd
+    Ok(host::__wasi_filestat_t {
+        st_dev: hostcalls_impl::device_id(file)?,
+        st_ino: hostcalls_impl::file_serial_no(file)?,
+        st_nlink: hostcalls_impl::num_hardlinks(file)?
+            .try_into()
+            .expect("overflow"),
```
kubkon: Here, we should probably return `host::__WASI_EOVERFLOW`.

src/sys/windows/hostcalls_impl/fs.rs
```diff
@@ -173,8 +173,39 @@ pub(crate) fn path_rename(
...
+    let low = info.nFileIndexLow;
+    let no = ((high as u64) << 32) | (low as u64);
+    Ok(no)
+}
+
+pub(crate) fn change_time(file: &File) -> io::Result<u64> {
+    winx::file::change_time(file)
+}
+
+pub(crate) fn filetype(file: &File) -> io::Result<host::__wasi_filetype_t> {
```
kubkon: I wonder, should we perhaps augment this function with `winx::file::get_file_type` for now? This function allows you to match for `__WASI_FILETYPE_CHARACTER_DEVICE` and `__WASI_FILETYPE_SOCKET_STREAM`. Also, now that I look at this function here, it would be good if we could re-use in `sys::windows::fdentry_impl::determine_type_rights`.

src/sys/unix/hostcalls_impl/fs.rs
```diff
@@ -356,11 +357,49 @@ pub(crate) fn path_rename(
...
+    use std::os::unix::fs::MetadataExt;
+    Ok(file.metadata()?.dev())
+}
+
+pub(crate) fn file_serial_no(file: &File) -> io::Result<u64> {
+    use std::os::unix::fs::MetadataExt;
+    Ok(file.metadata()?.ino())
+}
+
+pub(crate) fn filetype(file: &File) -> io::Result<host::__wasi_filetype_t> {
```
kubkon: Could we re-use this function in `sys::unix::fdentry_impl::determine_type_rights`?

winx/src/file.rs
```diff
@@ -457,3 +459,48 @@ pub fn openat<S: AsRef<OsStr>>(
         Ok(handle)
     }
 }
+
+fn cvt(i: winapi::shared::minwindef::BOOL) -> io::Result<()> {
```
kubkon, line 464: Two things here:
1) could you reapply this optimisation to the rest of functions in `winx`?
2) could we make this more generic to accept any type and return a type?

src/sys/windows/hostcalls_impl/fs.rs
kubkon: Then again, it might not make that much sense if we are specifically expecting a `File` especially on Windows 🤔

kubkon:
cc @sunfishcode @alexcrichton

build.rs
kubkon: Alternatively, could you actually try removing `cleanup_file` call from `file_allocate` testcase? You can find the source code [here](https://github.com/CraneStation/wasi-misc-tests).

src/hostcalls_impl/fs.rs
marmistrz: I'm returning `io::Result` in `fd_filestat_get_impl`, so I can't. I could probably introduce a new enum to contain either the WASI error or `std::io::Error`. Or add a convenience helper function to convert `std::io::Error` to the WASI error so that I can just `map_err`

winx/src/file.rs
marmistrz: 1. I'd prefer to make it a subsequent PR
2. Yes, but I'll probably need to depend on `num-traits` or make my own custom trait (the latter is what libstd does, this is basically copy-paste from libstd, so I should probably add the comment about it)

build.rs
marmistrz: Yes, the test succeeds if I remove `cleanup_file`.

winx/src/file.rs
kubkon: That’s fair enough. Can you add a comment here then?

marmistrz: I've already done it.

build.rs
kubkon: Can you include a modified `file_allocate.wasm` binary here and submit a PR with the required modifications to `wasi-misc-tests`?

kubkon: Ahh, I see you've reverted the change already. That's fine then :-)

src/hostcalls_impl/fs.rs
kubkon: But could you not return our `Result` type immediately without introducing a new enum nor a convenience helper? Then, I feel like we would retain finer grain control over the error values. For example, with the current approach, regardless of where the error happens along the way, we'll return a `io::ErrorKind::Other` which will get mapped to `host::__WASI_EIO`, whereas if had our `Result` type returned, then `st.duration_since(..)` could return an `host::__WASI_EINVAL` while later on, on `try_into()` you'd map to `host::__WASI_EILSEQ`. Also, I see that only `file.metadata()` returns an `io::Result`, and your helper fns could be rewritten to return our `Result` type for now.

kubkon: I agree that in a future PR we should refactor all of the `map_err` by introducing another enum type or something, but one step at a time :-)

marmistrz: @kubkon  most of the `libstd` function return `io::Result` and I'd like to (1) avoid repeating `         e.raw_os_error().map_or(host::__WASI_EIO, errno_from_host)` over and over (2) be able to `debug!` print the `io::Error` in case it's `ErrorKind::Other` (sometimes returned by `libstd`).

`io::Error` is returned by: metadata, accessed, modified. `metadata` is also extensively used by all the helper functions on Unix, when using Windows syscalls I also return an `io::Error` from last OS error.

On the other hand, all the wasi errors appear in the platform-independent `fd_filestat_get_impl`. Hence, IMO the platform dependent code should return `io::Result` and the platform-independent code should convert them to wasi error types

src/sys/unix/hostcalls_impl/fs.rs
```diff
@@ -356,11 +357,49 @@ pub(crate) fn path_rename(
...
 }
 
-pub(crate) fn fd_filestat_get(fd: &File) -> Result<host::__wasi_filestat_t> {
-    use nix::sys::stat::fstat;
-    let filestat =
-        fstat(fd.as_raw_fd()).map_err(|err| host_impl::errno_from_nix(err.as_errno().unwrap()))?;
-    host_impl::filestat_from_nix(filestat)
+pub(crate) fn num_hardlinks(file: &File) -> io::Result<u64> {
+    use std::os::unix::fs::MetadataExt;
+    Ok(file.metadata()?.nlink())
```
alexcrichton: This looks like it re-executes `stat` syscalls for each of these functions, but this should probably operate on `&fs::Metadata` to avoid a separate syscall?

src/sys/windows/hostcalls_impl/fs.rs
```diff
@@ -173,8 +173,39 @@ pub(crate) fn path_rename(
     unimplemented!("path_rename")
 }
 
-pub(crate) fn fd_filestat_get(fd: &File) -> Result<host::__wasi_filestat_t> {
-    unimplemented!("fd_filestat_get")
+pub(crate) fn num_hardlinks(file: &File) -> io::Result<u64> {
+    Ok(winx::file::get_fileinfo(file)?.nNumberOfLinks.into())
```
alexcrichton, line 179: Sort of like the Unix implementation, could this operate on `fs::Metadata` or some similar structure which is only created once instead of being created multiple times?

src/hostcalls_impl/fs.rs
sunfishcode: I share @kubkon's concern about retaining fine-grained error values. Rust's [`ErrorKind`](https://doc.rust-lang.org/std/io/enum.ErrorKind.html) doesn't have values for all the error codes that we use, such as `EOVERFLOW` and `EILSEQ`.

In the case of `duration_since`, if I read the documentation correctly, it'll fail here if given a time before the UNIX epoch. My interpretation is that `EOVERFLOW` is the most appropriate error code for this; eg. [POSIX says](http://pubs.opengroup.org/onlinepubs/9699919799/) "[EOVERFLOW]
    A value to be stored would overflow one of the members of the stat structure." So we'd like to be able to return that rather than `EIO`.

src/sys/unix/hostcalls_impl/fs.rs
marmistrz: I'd love to operate on `&fs::Metadata` but then it'd be impossible to implement those helper functions on Windows. While the Rust libstd actually does the desired syscall (`GetFileInformationByHandle`), it simply throws away some of the information, such as `nNumberOfLinks` or other fields.
We'd have to extend `io::FileAttr` on Windows so that we can operate on metadata and not on `&File`. 
https://github.com/rust-lang/rust/blob/master/src/libstd/sys/windows/fs.rs#L146

As described in the OP, it was a design decision to accept superfluous syscalls. Alternatively I can alternatively combine all those helper functions into a single one.

After we've merged this PR, I'd like to discuss with you the possibilities of extending the Rust `libstd` so that we can eventually use `&fs::Metadata` for the helper functions (when the changes are stabilized)

marmistrz:
I modified the code to return proper error codes. But now I'm wondering if it makes any sense to have a separate `fd_filestat_get_impl` - I made it a separate function because it was meant to return an `io::Error`

src/sys/unix/hostcalls_impl/fs.rs
alexcrichton: Yes that's true that rust libstd throws it away (and we should fix that) but there's no need to issue new stat calls on Linux. I think it would be a pretty easy refactor to pass in both `&File` and `&Metadata` here and provide some ability to cache information between these field accesors.

marmistrz: Fair enough, I'll do as you suggest.

alexcrichton:
Looking more at this I think it may be best actually to have the original organization where there's a totally platform-specific implementation for both Windows and non-Windows. On non-Windows it can be reimplemented to use `file.metadata()`, but on Windows I think it probably wants to execute `GetFileInformationByHandle` only once and then use information from there. 

Looking at Rust's libstd there's two ways to acquire a `fs::Metadata`, one from `DirEntry` and the other from `fs::metadata`. The two sources have different bits of data on them and `fs::Metadata` is the intersection of the data. We probably want to add an extension trait for accessing `fs::Metadata` extra information when it's learned through `fs::Metadata` to avoid having to reproduce all of the information here.

marmistrz:
You're right. 
`fs::metadata` has information taken from `GetFileInformationByHandle` and `DirEntry` from `FindNextFileW`. 

I'll do as you suggest.

> Looking at Rust's libstd there's two ways to acquire a fs::Metadata, one from DirEntry and the other from fs::metadata. The two sources have different bits of data on them and fs::Metadata is the intersection of the data. We probably want to add an extension trait for accessing fs::Metadata extra information when it's learned through fs::Metadata to avoid having to reproduce all of the information here.

We could extend `std::os::windows::fs::MetadataExt` and return an `Option<_>`, and switch to the extension of `MetadataExt` when it's stabilized

This will still need two syscalls on Windows: `GetFileInformationByHandle` once & `GetFileInformationByHandleEx` once but I doubt we can avoid it. I'll recheck tomorrow if we can (UTC+2)

alexcrichton:
@marmistrz [my thoughts exactly of how to do this](https://github.com/rust-lang/rust/pull/62980) :)

marmistrz:
@alexcrichton I thought I was going to have my first `libstd` pull request :P 
LGTM :)

alexcrichton:
Oh oops my bad! I certainly don't want to steal any thunder by any means

kubkon:
Hahaha, @alexcrichton @marmistrz love the banter! You guys are the best :-D

kubkon:
LGTM! @sunfishcode @marmistrz do you guys have any other comments you'd like addressed?

marmistrz:
Could we extend
```
pub type __wasi_linkcount_t = u32;
``` 
to `u64` so that there is no overflow possible in `nlink()`?

kubkon:
@marmistrz AFAIK that might not be that simple as it’d require changing the WASI spec. I’m sure that @sunfishcode will be able to shed some light on this.

marmistrz:
I rebased and refactored the change as @alexcrichton suggested. I left over a duplicate `convert_err` helper function for now. 
I can already see I'm going to use such a helper function a lot (I need it for my current draft of `fd_filestat_set_size`, it would also be useful if we implemented `fd_filestat_set_times` using the platform-agnostic `filetime` crate), so I have the following ideas, in increasing order of complexity.

1. Add `errno_from_io_error` which is `std::io::Error -> host::__wasi_errno_t` to `crate::sys` which would correspond to the current `convert_err`. In particular, I find it important that we print `debug!` information if some information about the error is lost.
2. Create a new type 
```
enum Error {
     RawError(host::__wasi_errno_t),
     OsError(std::io::Error),
}
```
add a conversion method and handle the conversion as a part of the `hostcalls` macro. Also, `From<std::io::Error>`
3. As in two, but swap the inner `host::__wasi_errno_t` for a new type, say, `WASIError` which would implement `TryFrom<host::__wasi_errno_t>`. Then it would be possible to have `impl From<WASIError> for Error`

2 or 3 would have to be done as a subsequent pull request. What do you think?

marmistrz:
The CI failure is unrelated to the changes in this PR, master is currently failing.

kubkon:
> The CI failure is unrelated to the changes in this PR, master is currently failing.

I'm on it, it should be fixed in #45.

kubkon:
> I rebased and refactored the change as @alexcrichton suggested. I left over a duplicate `convert_err` helper function for now.
> I can already see I'm going to use such a helper function a lot (I need it for my current draft of `fd_filestat_set_size`, it would also be useful if we implemented `fd_filestat_set_times` using the platform-agnostic `filetime` crate), so I have the following ideas, in increasing order of complexity.
> 
>     1. Add `errno_from_io_error` which is `std::io::Error -> host::__wasi_errno_t` to `crate::sys` which would correspond to the current `convert_err`. In particular, I find it important that we print `debug!` information if some information about the error is lost.
> 
>     2. Create a new type
> 
> 
> ```
> enum Error {
>      RawError(host::__wasi_errno_t),
>      OsError(std::io::Error),
> }
> ```
> 
> add a conversion method and handle the conversion as a part of the `hostcalls` macro. Also, `From<std::io::Error>`
> 3. As in two, but swap the inner `host::__wasi_errno_t` for a new type, say, `WASIError` which would implement `TryFrom<host::__wasi_errno_t>`. Then it would be possible to have `impl From<WASIError> for Error`
> 
> 2 or 3 would have to be done as a subsequent pull request. What do you think?

Awesome, thanks! I like the way this is going, and temporarily I'd focus on option 1. for now until we land a working Windows implementation, which for me (can't speak for anyone else here) is a must-have and a priority. Afterwards, I'm happy to refactor in the most efficient/elegant way possible :-)

marmistrz:
I refactored the error handling, this should be ready. In a subsequent PR I'll reuse the `errno_from_ioerror` as much as possible so as to remove the code duplication.

kubkon:
I've now merged a fix for the CI in #45 so if you could rebase and check that everything tests out fine for you :-)

marmistrz:
Rebased, thanks

sunfishcode:
@marmistrz Yes, extending `__wasi_linkcount_t` to 64-bit makes sense to me, though we will indeed need a spec change, and it'll be an ABI change, both of which we can do, it'll just take some coordination. I filed https://github.com/WebAssembly/WASI/issues/70 to track this.

alexcrichton (approved):

PR #43, opened 2019-07-25
Mark all functions in host_impl.rs as pub(crate)

marmistrz:
Motivation for this PR: while working on #42 I wasn't sure if I hadn't forgotten about `filestat_from_nix`. If it were `pub(crate)`, the compiler would take care about it instead of me.


kubkon (approved):
LGTM!

PR #46, opened 2019-07-26
 Reuse errno_from_ioerror to simplify error handling 

marmistrz:
This PR bases on #42.

PR #47, opened 2019-07-26
 Implement fd_filestat_set_size using libstd 

marmistrz:
This PR bases on #42 & #46.

### DIFF
--- a/src/hostcalls_impl/fs.rs
+++ b/src/hostcalls_impl/fs.rs
@@ -440,6 +440,7 @@ pub(crate) fn fd_allocate(
 
     let current_size = metadata.len();
     let wanted_size = offset.checked_add(len).ok_or(host::__WASI_E2BIG)?;
+    // This check will be unnecessary when rust-lang/rust#63326 is fixed
     if wanted_size > i64::max_value() as u64 {
         return Err(host::__WASI_E2BIG);
     }
@@ -779,6 +780,10 @@ pub(crate) fn fd_filestat_set_size(
         .and_then(|fe| fe.fd_object.descriptor.as_file())?;
 
     let st_size = dec_filesize(st_size);
+    // This check will be unnecessary when rust-lang/rust#63326 is fixed
+    if st_size > i64::max_value() as u64 {
+        return Err(host::__WASI_E2BIG);
+    }
     fd.set_len(st_size).map_err(errno_from_ioerror)
 }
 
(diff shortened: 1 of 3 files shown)


src/hostcalls_impl/fs.rs
```diff
@@ -766,11 +766,7 @@ pub(crate) fn fd_filestat_set_size(
         .and_then(|fe| fe.fd_object.descriptor.as_file())?;
 
     let st_size = dec_filesize(st_size);
-    if st_size > i64::max_value() as u64 {
-        return Err(host::__WASI_E2BIG);
```
kubkon: Hmm, are you sure it is wise to remove this check? At a quick glance, `std::fs::File::set_len` which internally calls `std::sys::*::fs::File::truncate` does not perform this check. I might be mistaken here though! Please correct me if that's the case. If I'm right, however, I think we should leave this check in the code for better error signalling to the Wasm app. @sunfishcode @alexcrichton thoughts?

marmistrz: You're right: https://github.com/rust-lang/rust/blob/master/src/libstd/sys/unix/fs.rs#L552-L559
But isn't it a bug in libstd that a silent overflow occurs instead of an error being reported? (`off64_t` is `int64_t` on my system)

alexcrichton: Yes if libstd isn't doing a proper cast check then libstd should be where this error is generated for sure, sounds like a bug!

kubkon: Could we then leave the check as is for now and in the meantime, submit a bug and a patch to `libstd`? Sounds like the best option to me in the short term :-)

marmistrz: I'll submit the bug and make the patch to libstd tomorrow :)

kubkon: Awesome, thanks! :-)

kubkon: @marmistrz after you submit the patch and bring back the check in `wasi-common`, could you add a comment with a link to the `libstd`’s patch explaining that when it lands and stabilises we’ll remove it altogether?

marmistrz: Done.

marmistrz:
I rebased the change upon the new revision of #46.

kubkon:
Could you rebase onto master? Otherwise, the commit this PR is based on will be repeated again after we merge it seems :-)

marmistrz:
Done.

kubkon:
LGTM!

PR #49, opened 2019-08-02
Add azure pipelines definition

kubkon:
Thanks to @tschneidereit and his awesome PR CraneStation/wasmtime#237, I'm happy to announce that Azure Pipelines has landed!

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,0 +1,60 @@
+name: $(Build.SourceBranch)-$(date:yyyyMMdd)$(rev:.r)
+trigger:
+  branches:
+    include:
+    - '*'
+
+jobs:
+- job: Build
+  strategy:
+    matrix:
+      windows-stable:
+        imageName: 'vs2017-win2016'
+        rustup_toolchain: stable
+      linux-stable:
+        imageName: 'ubuntu-16.04'
+        rustup_toolchain: stable
+      mac-stable:
+        imageName: 'macos-10.14'
+        rustup_toolchain: stable
+
+  pool:
+    vmImage: $(imageName)
+
+  steps:
+    - checkout: self
+    - script: |
+        curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain $RUSTUP_TOOLCHAIN
+        echo "##vso[task.setvariable variable=PATH;]$PATH:$HOME/.cargo/bin"
+        export PATH=$PATH:$HOME/.cargo/bin
+        rustup update $RUSTUP_TOOLCHAIN
+        rustup default $RUSTUP_TOOLCHAIN
+        rustc --version
+      displayName: Install rust (*nix)
+      condition: ne( variables['Agent.OS'], 'Windows_NT' )
+    - script: |
+        rustup update %RUSTUP_TOOLCHAIN%
+        rustup default %RUSTUP_TOOLCHAIN%
+        rustc --version
+      displayName: Install rust (Win)
+      condition: eq( variables['Agent.OS'], 'Windows_NT' )
+    - script: |
+        rustup component add --toolchain stable rustfmt-preview
+        rustup component list --toolchain stable
+      displayName: Add rustfmt
+    - script: |
+        powershell -Command "$ProgressPreference = 'SilentlyContinue'; iwr -outf %TEMP%\LLVM-8.0.0-win64.exe https://rust-lang-ci2.s3.amazonaws.com/rust-ci-mirror/LLVM-8.0.0-win64.exe"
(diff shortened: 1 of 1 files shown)


kubkon:
cc @tschneidereit

kubkon:
@tschneidereit could I ask you to set up Azure for the repo please, and turn off Travis? After this is merged, I'll fix badges in the README to point to Azure :-)

.azure-pipelines.yml
```diff
@@ -0,0 +1,60 @@
...
+        rustup component list --toolchain stable
+      displayName: Add rustfmt
+    - script: |
+        powershell -Command "$ProgressPreference = 'SilentlyContinue'; iwr -outf %TEMP%\LLVM-8.0.0-win64.exe https://rust-lang-ci2.s3.amazonaws.com/rust-ci-mirror/LLVM-8.0.0-win64.exe"
+        %TEMP%\LLVM-8.0.0-win64.exe /S /NCRC
+      condition: eq(variables['Agent.OS'], 'Windows_NT')
+      displayName: Install clang (Win)
+    - script: cargo fetch
+      displayName: Fetch cargo dependencies
+    - script: cargo +stable fmt --all -- --check
```
tschneidereit, line 52: the `+stable` shouldn't be needed here. It also really doesn't matter, of course :)

.azure-pipelines.yml
```diff
@@ -0,0 +1,60 @@
...
+      displayName: Check formatting
+    - script: cargo build --release
+      displayName: Cargo build release
+    - script: cargo build
+      displayName: Cargo build
```
tschneidereit, line 57: The reason I split up the debug and release build configs in the Wasmtime config is that it significantly reduces wall-clock time for the test run overall. How long do the build phases take for wasi—might it be worth splitting them up in the same way?

tschneidereit (approved):
Looks great! I'm entirely fine with landing this as-is, and potentially addressing the issues I mentioned as follow-ups, if at all.

.azure-pipelines.yml
kubkon: Mhm, that's what I thought and rightly done :-) In this case however, both `cargo build` and `cargo build --release` are not the bottlenecks (the former takes around 1min while the latter 3min), it's `cargo test --all` that can take much longer (and indeed took forever on Win in Travis!). In Azure, tests take from 5 to 6mins roughly (depending on the host with Win being the slowest but still much better performance than in Travis which is just great!).

.azure-pipelines.yml
kubkon: Good catch!

tschneidereit:
I'll merge this now so I can finish setting up the pipeline.

tschneidereit:
Running here: https://dev.azure.com/CraneStation/wasi-common/_build/results?buildId=12

kubkon:
SGTM! We can always expand the config in the future should we for example decide to include `beta` and `nightly` builds (although for now I'm guessing they are not needed).

kubkon:
> Running here: https://dev.azure.com/CraneStation/wasi-common/_build/results?buildId=12

Rock on! 🎸 👩‍🎤 🎤

.azure-pipelines.yml
```diff
@@ -0,0 +1,60 @@
...
+      displayName: Install rust (*nix)
+      condition: ne( variables['Agent.OS'], 'Windows_NT' )
+    - script: |
+        rustup update %RUSTUP_TOOLCHAIN%
+        rustup default %RUSTUP_TOOLCHAIN%
+        rustc --version
+      displayName: Install rust (Win)
+      condition: eq( variables['Agent.OS'], 'Windows_NT' )
+    - script: |
+        rustup component add --toolchain stable rustfmt-preview
```
tschneidereit, line 42: One thing I just realize: this might break if we introduce toolchains other than stable as configurations. The Install rust steps above will probably not install the stable toolchain in that case.

kubkon: Ah yes, well spotted! I'll fix it in a subsequent PR and cc you in :-)

.azure-pipelines.yml
tschneidereit: Makes sense, yes. You'd still save 1 minute of wall-clock time, or about 10%, because the builds would be parallelized—plus you'd get test coverage for release builds (with some additional boilerplate, admittedly)

kubkon: Yep, that's true. I'll leave it as-is for now, but we can split it like this in the future. In case of `wasmtime` it makes a lot more sense since we're not only building a lib but also binaries, whereas here, only lib.

tschneidereit: Agreed, and sounds good to me!

PR #51, opened 2019-08-02
Remove Travis conf and redundant scripts

kubkon:


test-all.sh
```diff
@@ -1,53 +0,0 @@
...
-banner "Rust debug build"
-cargo build
-
-# Run the tests. We run these in debug mode so that assertions are enabled.
-banner "Rust unit and integration tests"
-RUST_BACKTRACE=1 cargo test --all
-
-# Make sure the documentation builds.
-banner "Rust documentation: $topdir/target/doc/wasi-common/index.html"
-cargo doc
```
tschneidereit, line 51: I guess we've lost the `cargo doc` invocation. Same for the Wasmtime repo, where I'm adding it in right now.

tschneidereit (approved):
LGTM!

test-all.sh
kubkon: Ahh, well spotted! I'll do the same for wasi-common :-)

PR #52, opened 2019-08-02
Add cargo doc and apply cargo fmt to stable only

kubkon:


kubkon:
cc @tschneidereit

.azure-pipelines.yml
```diff
@@ -38,23 +38,25 @@ jobs:
...
+      condition: eq( variables['rustup_toolchain'], 'stable' )
       displayName: Check formatting
     - script: cargo build --release
       displayName: Cargo build release
     - script: cargo build
       displayName: Cargo build
     - bash: RUST_BACKTRACE=1 cargo test --all
       displayName: Cargo test
+    - script: cargo doc
+      displayName: Cargo doc
```
tschneidereit, line 62: Do docs behave differently based on the platform? I.e., are docs for functions only included if the platform is targeted? If not, it might make sense to only run this on macOS, because the builders for that are fastest.

.azure-pipelines.yml
```diff
@@ -38,23 +38,25 @@ jobs:
...
```
tschneidereit: Pre-existing nit: superfluous newline :)

tschneidereit (approved):
Looks great! Feel free to merge, or push a follow-up for the one comment.

.azure-pipelines.yml
kubkon: Cool suggestion, thanks! I didn't know that macOS builders are the fastest in Azure, and no, I don't think they should behave differently based on the platform.

PR #55, opened 2019-08-06
Make fd_write unbuffered; fixes CraneStation/wasmtime#255

### LINKED ISSUE #255: fd_write should be unbuffered
fd_write should be unbuffered to match the behavior of writev. Buffering can be implemented in the application if desired, however, fd_write being buffered means there is no way to do unbuffered writes (there is no WASI flush API call and there shouldn't be at this level). This is using the dev release of wasmtime from today (95bcc63). This is a regression from a a few months ago. Note that recent builds of wasmer and lucet have unbuffered fd_write.

Here is a simple test case that should print a prompt, read a string and echo it:

```
(module
  (import "wasi_unstable" "fd_write" (func $fd_write (param i32 i32 i32 i32) (result i32)))
  (import "wasi_unstable" "fd_read" (func $fd_read (param i32 i32 i32 i32) (result i32)))

  (memory 10)
  (export "memory" (memory 0))

  (func $main (local i32)
    (i32.store offset=0 (i32.const 4) (i32.const 12))  ;; iovec ptr
    (i32.store offset=4 (i32.const 4) (i32.const 2))   ;; iovec len
    (i32.store offset=0 (i32.const 12) (i32.const 62)) ;; string '> '
    (i32.store offset=1 (i32.const 12) (i32.const 32))
    (i32.store offset=2 (i32.const 12) (i32.const 0))
    ;; print prompt
    (drop (call $fd_write (i32.const 1) (i32.const 4) (i32.const 1) (i32.const 256)))

    (i32.store offset=0 (i32.const 16) (i32.const 24))  ;; iovec ptr
    (i32.store offset=4 (i32.const 16) (i32.const 200)) ;; iovec len
    ;; read
    (drop (call $fd_read (i32.const 0) (i32.const 16) (i32.const 1) (i32.const 256)))
    ;; echo
    (drop (call $fd_write (i32.const 1) (i32.const 16) (i32.const 1) (i32.const 256)))
  )

  ;;(start $main)
  (export "_start" $main)
)
```

And here is the behavior:

```
$ wasm-as wasitests/echo2.wat -o wasitests/echo2.wasm
$ wasmtime wasitests/echo2.wasm
typed by the user
> typed by the user
```

The correct behavior should be:

```
$ wasmtime wasitests/echo2.wasm
> typed by the user
typed by the user
```

kubkon:
This PR fixes CraneStation/wasmtime#255 by making `fd_write` unbuffered. AFAIK, `std::io::Read` and `std::io::Write` all do buffered reads and writes respectively. I'm not sure whether the fact that our implementation of `fd_read` uses buffered reads is a problem?

@marmistrz I've not re-used your fn `errno_from_ioerror` as I didn't want to create additional conflicts for your PR #46.

### DIFF
--- a/src/hostcalls_impl/fs.rs
+++ b/src/hostcalls_impl/fs.rs
@@ -376,15 +376,29 @@ pub(crate) fn fd_write(
         .map(|vec| unsafe { host::iovec_to_host(vec) })
         .collect();
 
-    let maybe_host_nwritten = match &mut *fe.fd_object.descriptor {
-        Descriptor::File(f) => f.write_vectored(&iovs),
+    // perform unbuffered writes
+    let host_nwritten = match &mut *fe.fd_object.descriptor {
+        Descriptor::File(f) => f
+            .write_vectored(&iovs)
+            .map_err(|err| err.raw_os_error().map_or(host::__WASI_EIO, errno_from_host))?,
         Descriptor::Stdin => return Err(host::__WASI_EBADF),
-        Descriptor::Stdout => io::stdout().lock().write_vectored(&iovs),
-        Descriptor::Stderr => io::stderr().lock().write_vectored(&iovs),
+        Descriptor::Stdout => {
+            // lock for the duration of the scope
+            let stdout = io::stdout();
+            let mut stdout = stdout.lock();
+            let nwritten = stdout
+                .write_vectored(&iovs)
+                .map_err(|err| err.raw_os_error().map_or(host::__WASI_EIO, errno_from_host))?;
+            stdout
+                .flush()
+                .map_err(|err| err.raw_os_error().map_or(host::__WASI_EIO, errno_from_host))?;
+            nwritten
+        }
+        Descriptor::Stderr => io::stderr()
+            .lock()
+            .write_vectored(&iovs)
+            .map_err(|err| err.raw_os_error().map_or(host::__WASI_EIO, errno_from_host))?,
     };
-
-    let host_nwritten = maybe_host_nwritten
-        .map_err(|err| err.raw_os_error().map_or(host::__WASI_EIO, errno_from_host))?;
 
     trace!("     | *nwritten={:?}", host_nwritten);
 


kubkon:
cc @alexcrichton @sunfishcode

src/hostcalls_impl/fs.rs
```diff
@@ -376,11 +376,35 @@ pub(crate) fn fd_write(
...
+            nwritten
+        }
+        Descriptor::Stderr => {
+            // lock for the duration of the scope
+            let stderr = io::stderr();
+            let mut stderr = stderr.lock();
+            let nwritten = stderr.write_vectored(&iovs);
+            stderr
+                .flush()
+                .map_err(|err| err.raw_os_error().map_or(host::__WASI_EIO, errno_from_host))?;
```
sunfishcode, line 394: Unlike `io::stdout`, `io::stderr` is [not buffered](https://doc.rust-lang.org/std/io/fn.stderr.html), so we don't need to flush it.

src/hostcalls_impl/fs.rs
```diff
@@ -376,11 +376,35 @@ pub(crate) fn fd_write(
         .map(|vec| unsafe { host::iovec_to_host(vec) })
         .collect();
 
+    // perform unbuffered writes
     let maybe_host_nwritten = match &mut *fe.fd_object.descriptor {
-        Descriptor::File(f) => f.write_vectored(&iovs),
+        Descriptor::File(f) => {
+            let nwritten = f.write_vectored(&iovs);
+            f.flush()
+                .map_err(|err| err.raw_os_error().map_or(host::__WASI_EIO, errno_from_host))?;
```
sunfishcode: The Rust documentation isn't clear about this, but if I read the implementation correctly, on [Windows](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/windows/handle.rs#L170) and [Unix](https://github.com/rust-lang/rust/blob/master/src/libstd/sys/unix/fd.rs#L120), `File`'s `write` functions don't do buffering, so we shouldn't need to flush them.

src/hostcalls_impl/fs.rs
```diff
@@ -376,11 +376,35 @@ pub(crate) fn fd_write(
...
-        Descriptor::Stdout => io::stdout().lock().write_vectored(&iovs),
-        Descriptor::Stderr => io::stderr().lock().write_vectored(&iovs),
+        Descriptor::Stdout => {
+            // lock for the duration of the scope
+            let stdout = io::stdout();
+            let mut stdout = stdout.lock();
+            let nwritten = stdout.write_vectored(&iovs);
+            stdout
+                .flush()
+                .map_err(|err| err.raw_os_error().map_or(host::__WASI_EIO, errno_from_host))?;
```
sunfishcode, line 391: This is really subtle, but the ordering here means that errors from the `flush` call are reported before errors from the `write_vectored` call. I admittedly can't think of a case where it would matter offhand, but it feels untidy to check errors out of order.

marmistrz:
@kubkon you can rebase onto #46 if you feel like, or I can also refactor your change later on.

src/hostcalls_impl/fs.rs
marmistrz: On the other hand, it may be a good idea to add the currently-no-op flush in case the Rust implementation decides to do some buffering. Or to clarify the docs.

src/hostcalls_impl/fs.rs
kubkon: Excellent point!

src/hostcalls_impl/fs.rs
kubkon: Ah, of course! Well spotted, thanks!

src/hostcalls_impl/fs.rs
kubkon: @sunfishcode I think you're spot on! @marmistrz I'm partial to leaving flushing out of this one, and clarifying Rust's docs instead. Then, if anything changes we should in principle be able to track it down through the changes in the docs. :-)

kubkon:
> @kubkon you can rebase onto #46 if you feel like, or I can also refactor your change later on.

I was thinking of not rebasing onto #46 as this PR is a hotfix rather than a major change, and then we could apply your refactoring #46 when we sync with master?

marmistrz:
@kubkon ok

src/hostcalls_impl/fs.rs
sunfishcode: Also, if Rust ever does decide to make `File`'s `write` etc. buffered, it'd be better to switch to a lower-level API without buffering, rather than use buffering and flushing.

It's unfortunate that we have to deal with stdout buffering, but it is useful to coordinate with `io::stdout` so that we interop with Rust code in the host.

sunfishcode:
Looks good!

PR #54, opened 2019-08-05
 Implement fd_filestat_set_times using the filetime crate. 

marmistrz:
In the future we could consider using `filetime` to replace `systemtime_to_timestamp` from #42.

PR #56, opened 2019-08-09
Add a testcase for fd_filestat_set_*

marmistrz:
This PR matches https://github.com/CraneStation/wasi-misc-tests/pull/14 and should be merged before #54.
The file was created using `cargo build --release --target=wasm32-wasi`

(I'm not really sure how I should submit the .wasm files)


build.rs
```diff
@@ -136,6 +136,7 @@ cfg_if::cfg_if! {
                     "symlink_loop" => true,
                     "clock_time_get" => true,
                     "truncation_rights" => true,
+                    "fd_filestat_set" => true,
```
kubkon, line 139: I'm at a loss here. I thought that the whole point was to add a testcase that tests changes of #54 in addition to #47  :-D Anyhow, I suggest we wait with this PR after #54 is merge, or you could potentially add the test binary as part of #54.

kubkon:
> This PR matches [CraneStation/wasi-misc-tests#14](https://github.com/CraneStation/wasi-misc-tests/pull/14) and should be merged before #54.
> The file was created using `cargo build --release --target=wasm32-wasi`
> 
> (I'm not really sure how I should submit the .wasm files)

I'd personally submit the testcase as part of the PR which introduces a new feature, so in this case, #54. But this way is good provided we merge #54 first and we don't exclude the test from Windows build ;-)

marmistrz:
The idea was to guarantee that we don't accidentally change the semantics of fd_filestat_set_times. Now we test only #47, we'll enable this test on Windows as a part of #54.

kubkon:
Ah, that's an excellent point actually the results of which we have witnessed first hand in CraneStation/wasmtime#255 ;-)

PR #60, opened 2019-08-10
Add path_rename implementation on Windows

kubkon:
This PR adds `path_rename` implementation on Windows. It depends on #57 to be merged first.

The implementation handles and is tested against all cases listed in CraneStation/wasi-misc-tests#17. However, checks that symlinks and links are handled properly are still missing, but I reckon we can do that in a future PR when we add a testcase testing such a scenario.

### DIFF
--- a/src/sys/windows/hostcalls_impl/fs.rs
+++ b/src/sys/windows/hostcalls_impl/fs.rs
@@ -213,26 +213,29 @@ pub(crate) fn path_rename(resolved_old: PathGet, resolved_new: PathGet) -> Resul
     if old_path.is_dir() && new_path.is_file() {
         return Err(host::__WASI_ENOTDIR);
     }
-    if old_path.is_file() && new_path.is_dir() {
-        return Err(host::__WASI_EISDIR);
-    }
 
     // TODO handle symlinks
 
     fs::rename(&old_path, &new_path).or_else(|e| match e.raw_os_error() {
         Some(e) => {
-            if old_path.is_dir() && new_path.is_dir() {
-                if !dir_is_empty(&new_path).map_err(errno_from_ioerror)? {
-                    Err(host::__WASI_ENOTEMPTY)
-                } else {
-                    // remove new_path dir and rename
-                    fs::remove_dir(&new_path)
-                        .and_then(|()| fs::rename(old_path, new_path))
-                        .map_err(errno_from_ioerror)
+            use winx::winerror::WinError;
+
+            log::debug!("path_rename at rename error code={:?}", e);
+            match WinError::from_u32(e as u32) {
+                WinError::ERROR_ACCESS_DENIED => {
+                    // So most likely dealing with new_path == dir.
+                    // Eliminate case old_path == file first.
+                    if old_path.is_file() {
+                        Err(host::__WASI_EISDIR)
+                    } else {
+                        // Ok, let's try removing an empty dir at new_path if it exists
+                        // and is a nonempty dir.
+                        fs::remove_dir(&new_path)
+                            .and_then(|()| fs::rename(old_path, new_path))
+                            .map_err(errno_from_ioerror)
+                    }
                 }
-            } else {
-                log::debug!("path_rename at rename error={:?}", e);
-                Err(errno_from_host(e))
+                e => Err(host_impl::errno_from_win(e)),
             }
         }
         None => {
@@ -240,14 +243,6 @@ pub(crate) fn path_rename(resolved_old: PathGet, resolved_new: PathGet) -> Resul
             Err(host::__WASI_EIO)
         }
     })
-}
-
-fn dir_is_empty<P: AsRef<Path>>(path: P) -> io::Result<bool> {
-    let mut it = std::fs::read_dir(path)?;
-    match it.next() {
-        Some(_) => Ok(false),
-        None => Ok(true),
-    }
 }
 
 pub(crate) fn num_hardlinks(file: &File, _metadata: &Metadata) -> io::Result<u64> {
--- a/src/sys/windows/hostcalls_impl/fs.rs
+++ b/src/sys/windows/hostcalls_impl/fs.rs
@@ -41,13 +41,11 @@ pub(crate) fn fd_pread(
     buf: &mut [u8],
     offset: host::__wasi_filesize_t,
 ) -> Result<usize> {
-    read_at(file, buf, offset)
-        .map_err(errno_from_ioerror)
+    read_at(file, buf, offset).map_err(errno_from_ioerror)
 }
 
 pub(crate) fn fd_pwrite(file: &File, buf: &[u8], offset: host::__wasi_filesize_t) -> Result<usize> {
-    write_at(file, buf, offset)
-        .map_err(errno_from_ioerror)
+    write_at(file, buf, offset).map_err(errno_from_ioerror)
 }
 
 pub(crate) fn fd_fdstat_get(fd: &File) -> Result<host::__wasi_fdflags_t> {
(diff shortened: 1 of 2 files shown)


src/sys/windows/hostcalls_impl/fs.rs
```diff
@@ -203,7 +201,53 @@ pub(crate) fn path_readlink(resolved: PathGet, buf: &mut [u8]) -> Result<usize>
 }
 
 pub(crate) fn path_rename(resolved_old: PathGet, resolved_new: PathGet) -> Result<()> {
-    unimplemented!("path_rename")
+    use std::fs;
+
+    let old_path = concatenate(resolved_old.dirfd(), Path::new(resolved_old.path()))?;
+    let new_path = concatenate(resolved_new.dirfd(), Path::new(resolved_new.path()))?;
```
sunfishcode, line 207: It looks like it'd be useful to factor out a function which takes a `PathGet` and returns a full path would be useful, here and in few pre-existing places.

src/sys/windows/hostcalls_impl/fs.rs
```diff
@@ -203,7 +201,53 @@ pub(crate) fn path_readlink(resolved: PathGet, buf: &mut [u8]) -> Result<usize>
...
+
+    // first, sanity check that we're not trying to rename dir to file or vice versa
+    // NB on Windows, the former is actually permitted [std::fs::rename]
```
sunfishcode: Please add a period here so that it's clear that this is meant to be two sentences.

src/sys/windows/hostcalls_impl/fs.rs
```diff
@@ -203,7 +201,53 @@ pub(crate) fn path_readlink(resolved: PathGet, buf: &mut [u8]) -> Result<usize>
...
+    //
+    // [std::fs::rename]: https://doc.rust-lang.org/std/fs/fn.rename.html
+    if old_path.is_dir() && new_path.is_file() {
+        return Err(host::__WASI_ENOTDIR);
+    }
+    if old_path.is_file() && new_path.is_dir() {
+        return Err(host::__WASI_EISDIR);
+    }
```
sunfishcode: These checks can race with other processes which could potentially be creating or deleting files or directories named `new_path` or `old_path` between when these checks and when we actually do the `fs::rename`. In general, the way to avoid these kinds of races is to just do the operation, `fs::rename` in this case, without doing any checks first, and then handling failures in the error cases.

Does `rename` return error codes that allow us to return `EISDIR`/`ENOTDIR` as needed?

src/sys/windows/hostcalls_impl/fs.rs
```diff
@@ -203,7 +201,53 @@ pub(crate) fn path_readlink(resolved: PathGet, buf: &mut [u8]) -> Result<usize>
...
+    fs::rename(&old_path, &new_path).or_else(|e| match e.raw_os_error() {
+        Some(e) => {
+            if old_path.is_dir() && new_path.is_dir() {
+                if !dir_is_empty(&new_path).map_err(errno_from_ioerror)? {
+                    Err(host::__WASI_ENOTEMPTY)
+                } else {
+                    // remove new_path dir and rename
+                    fs::remove_dir(&new_path)
+                        .and_then(|()| fs::rename(old_path, new_path))
+                        .map_err(errno_from_ioerror)
```
sunfishcode: Similarly, instead of explicitly checking for `new_path` being `is_dir`, it'd be better to just do `remove_dir` on it here, as that should fail if it's not a directory.

src/sys/windows/hostcalls_impl/fs.rs
kubkon: Excellent point!

src/sys/windows/hostcalls_impl/fs.rs
kubkon: Ah, that's a good point. So the problem here is that `std::fs::rename` on Windows does not error out in case when `old_path` is a ~~file~~ dir and `new_path` is a ~~dir~~ file, and according to WASI spec it should.

src/sys/windows/hostcalls_impl/fs.rs
kubkon: Will you mind if I refactor it out in a subsequent PR?

sunfishcode: Sounds good!

src/sys/windows/hostcalls_impl/fs.rs
sunfishcode: Aha, I missed that when reading this the first time. Is this an `fs::rename` behavior? If so, would switching to the underlying OS API help?

kubkon:
@sunfishcode a quick debugging session reveals the following issues with trying to be as atomic with syscalls as possible:
* `old_path == dir` and `new_path == file` is **not** an error on Windows
* for any combination where `new_path == dir`, the native OS error returned is always `ERROR_ACCESS_DENIED` which is a bummer

So, when it comes to handling the latter, I could potentially do it after trying `std::fs::rename` but then I'd still be forced to do `is_dir` checks to work out which case we're trying to handle. The former however requires an upfront check as otherwise `std::fs::rename` actually completes the operation. Thoughts? @alexcrichton would you be able to weigh in here as well please?

src/sys/windows/hostcalls_impl/fs.rs
kubkon: Rust seems not to do much except for wrap the native call. So this is pretty much verbatim OS API here: [MoveFileExW](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexw) with `MOVEFILE_REPLACE_EXISTING` flag on.

src/sys/windows/hostcalls_impl/fs.rs
kubkon: Hmm, this should potentially be doable. Let me have a closer look.

kubkon:
@sunfishcode OK, I've shuffled things around a bit now. I couldn't get rid of all the checks unfortunately (mainly due to weird error signalling by Windows API, and the fact that renaming anything to a file seems **not** to be an error), but, nonetheless, here it is. Could you have a look and let me know what you think?

kubkon:
Here’s another thought. What if we don’t rely on subsequent calls to `std::path::Path::is_dir` and related, but instead, at the beginning of the hostcall, we make a call to `std::fs::metadata` only once which then becomes a snapshot of info about a descriptor, and use that to make all the necessary assertions? Would that help @sunfishcode?

alexcrichton:
The Rust standard library [passes a flag](https://github.com/rust-lang/rust/blob/c01be67ea40266d6a4c3289654a07ddd7ce2a172/src/libstd/sys/windows/fs.rs#L657-L664) to allow overwriting, and that was done to mirror the Unix semantics of the `rename` function. I think though that `renameat` has the same behavior of overwriting on Unix as `rename` does, so I'd expect the wasi rename API to inherit that.

Does Unix end up actually returning an error though if you rename something and the destination already exists?

kubkon:
@alexcrichton excellent question! And the short answer is, that depends. Have a look at CraneStation/wasi-misc-tests#17 for a test matrix with all cases listed (excluding symlinks which I haven’t thought through yet).

EDIT: forgot to mention that current *nix implementation passes the test with flying colours: CraneStation/wasi-common#57

alexcrichton:
Hm sorry I'm not at a Windows computer to verify, but my reading of [the Windows documentation](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexa) seems to imply that file-to-file transfers are the same on Windows/Unix. Directory-to-file (and vice versa) transfers are errors on both platforms. Directory-to-directory renames look like they're inconsistent where on Unix it succeeds and the documentation indicates on Windows that it fails.

Is that perhaps the only case that's different here?

kubkon:
@alexcrichton Hmm, dir-to-file is not an error on Windows the last time I checked.

EDIT: I've temporarily disabled the upfront check for the case dir-to-file, and the test fails on Windows at trying to rename dir-to-file with the given status of `__WASI_ESUCCESS` whereas the expected status is an error `__WASI_ENOTDIR`. Unless I've really messed somewhere here (and if I did, please do correct me!), it seems that Windows does allow renaming dir-to-file after all. Here's the [link to assertion that panics](https://github.com/CraneStation/wasi-misc-tests/pull/17/files#diff-4cfece18a1438f951e87f5979c8b4e4bR131), and here's the [link to the CI output](https://dev.azure.com/CraneStation/wasi-common/_build/results?buildId=207).

alexcrichton:
Hm ok, well in any case I don't think this is something that's just well specified right now. It's basically "whatever Unix `rename` does" for the spec currently, and it's not clear if that's what we want long term. In the meantime I'm not sure it's possible to emulate the behavior on Windows in a non-racy fashion, so getting something working and then adding a fixme for later seems reasonable.

kubkon:
> Hm ok, well in any case I don't think this is something that's just well specified right now. It's basically "whatever Unix `rename` does" for the spec currently, and it's not clear if that's what we want long term. In the meantime I'm not sure it's possible to emulate the behavior on Windows in a non-racy fashion, so getting something working and then adding a fixme for later seems reasonable.

Hmm, @sunfishcode, what are your thoughts about all this? Should we proceed with the implementation as-is (minus retracting the last commit)?

sunfishcode:
Yeah, retracting the implementation as-is with the last commit retracted sounds good to me for now. We can bring this issue to the WASI Subgroup.

sunfishcode (approved):
Ok!

PR #61, opened 2019-08-13
Return EINVAL if both ATIM & ATIM_NOW, MTIM & MTIM_NOW are set in path_filestat_set_times

marmistrz:
For https://github.com/CraneStation/wasi-misc-tests/pull/19

### DIFF
--- a/src/hostcalls_impl/fs.rs
+++ b/src/hostcalls_impl/fs.rs
@@ -872,6 +872,15 @@ pub(crate) fn path_filestat_set_times(
         fst_flags
     );
 
+    let set_atim = fst_flags & host::__WASI_FILESTAT_SET_ATIM != 0;
+    let set_atim_now = fst_flags & host::__WASI_FILESTAT_SET_ATIM_NOW != 0;
+    let set_mtim = fst_flags & host::__WASI_FILESTAT_SET_MTIM != 0;
+    let set_mtim_now = fst_flags & host::__WASI_FILESTAT_SET_MTIM_NOW != 0;
+
+    if (set_atim && set_atim_now) || (set_mtim && set_mtim_now) {
+        return Err(host::__WASI_EINVAL);
+    }
+
     let dirfd = dec_fd(dirfd);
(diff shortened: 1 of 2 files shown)


src/hostcalls_impl/fs.rs
```diff
@@ -843,6 +843,12 @@ pub(crate) fn path_filestat_set_times(
         fst_flags
     );
 
+    if fst_flags & host::__WASI_FILESTAT_SET_ATIM & host::__WASI_FILESTAT_SET_ATIM_NOW != 0
+        || fst_flags & host::__WASI_FILESTAT_SET_MTIM & host::__WASI_FILESTAT_SET_MTIM_NOW != 0
```
sunfishcode: The logic here isn't what's intended; `fst_flags & host::__WASI_FILESTAT_SET_ATIM & host::__WASI_FILESTAT_SET_ATIM_NOW` always evaluates to zero.

marmistrz: What guarantees that they will never be set together? Nothing prevents me from setting `__wasi_fstflags_t` to `host::__WASI_FILESTAT_SET_ATIM & host::__WASI_FILESTAT_SET_ATIM_NOW` and this is what one of the tests in wasi-misc-tests does (for the fd_* version): https://github.com/CraneStation/wasi-misc-tests/pull/14

sunfishcode: `host::__WASI_FILESTAT_SET_ATIM & host::__WASI_FILESTAT_SET_ATIM_NOW` is zero. I agree that we need a check here, but we need the right check :-).

marmistrz: :D you're right!

marmistrz: fixed

marmistrz:
I rebased and moved the check to the Unix-specific implementation, because it will be unnecessary for the Windows implementation proposed in #63.

marmistrz:
Obsoleted by #64

PR #62, opened 2019-08-14
Miscellaneous refactoring

sunfishcode:
This just collects a few miscellanous refactorings I've collected while reviewing other changes.
 - Use char/byte literals instead of single-char/byte string literals.
 - Minor simplifications to `get_path_by_handle`.
 - Use `fs::read` instead of doing it manually.

### DIFF
--- a/tests/utils.rs
+++ b/tests/utils.rs
@@ -6,15 +6,8 @@ use std::io::prelude::*;
 use std::path::{Component, Path};
 use std::time::SystemTime;
 
-fn read_to_end<P: AsRef<Path>>(path: P) -> Result<Vec<u8>, io::Error> {
-    let mut buf: Vec<u8> = Vec::new();
-    let mut file = File::open(path)?;
-    file.read_to_end(&mut buf)?;
-    Ok(buf)
-}
-
 pub fn read_wasm<P: AsRef<Path>>(path: P) -> Result<Vec<u8>, String> {
-    let data = read_to_end(path).map_err(|err| err.to_string())?;
+    let data = fs::read(path).map_err(|err| err.to_string())?;
     if data.starts_with(&[b'\0', b'a', b's', b'm']) {
         Ok(data)
     } else {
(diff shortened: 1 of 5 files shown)


tests/utils.rs
```diff
@@ -6,15 +6,8 @@ use std::io::prelude::*;
...
 
-fn read_to_end<P: AsRef<Path>>(path: P) -> Result<Vec<u8>, io::Error> {
-    let mut buf: Vec<u8> = Vec::new();
-    let mut file = File::open(path)?;
-    file.read_to_end(&mut buf)?;
-    Ok(buf)
-}
-
 pub fn read_wasm<P: AsRef<Path>>(path: P) -> Result<Vec<u8>, String> {
-    let data = read_to_end(path).map_err(|err| err.to_string())?;
```
kubkon, line 17: We should probably backport this to `wasmtime` unless we already have :-)

winx/src/file.rs
```diff
@@ -360,8 +360,7 @@ pub fn get_file_access_mode(handle: RawHandle) -> Result<AccessMode> {
 pub fn get_path_by_handle(handle: RawHandle) -> Result<OsString> {
```
kubkon, line 360: This is not applicable to this PR, but I'm wondering, given the fact that we've narrowed down the scope of `winx` helper crate to basically two modules, `winerror` and `file`, should we perhaps move into `wasi-common` proper (inside `sys/windows` module)?

kubkon (approved):
LGTM!

winx/src/file.rs
sunfishcode: I don't have a strong opinion either way. On one hand, you're right, but on the other hand, `winx` could conceivably grow in the future to cover other things, as WASI evolves, or if we find uses for `winx` beyond `wasi-common`.

kubkon: That's a fair point. I guess it won't hurt if we simply leave as-is for now, and decide what to do with it in the future.

PR #63, opened 2019-08-14
 Implement path_filestat_get & path_filestat_set_times on Windows. 

marmistrz:
This still requires the test CraneStation/wasi-misc-tests#19


src/sys/windows/hostcalls_impl/fs.rs
```diff
@@ -320,15 +323,17 @@ pub(crate) fn path_filestat_set_times(
     mut st_mtim: host::__wasi_timestamp_t,
     fst_flags: host::__wasi_fstflags_t,
 ) -> Result<()> {
-    unimplemented!("path_filestat_set_times")
+    let path = resolved.concatenate()?;
+    let file = File::open(path).map_err(errno_from_ioerror)?;
+    fd_filestat_set_times_impl(&file, st_atim, st_mtim, fst_flags)
```
kubkon, line 332: I really like the fact you’re reusing the implementation between two hostcalls. I’m pretty sure we could do that in a couple more actually. It’ll be a good refactoring effort in a subsequent PR :-)

kubkon (approved):
LGTM! Thanks for refactoring out `concatenate` into a method of `PathGet` on Win. Looks great!

build.rs
```diff
@@ -136,6 +136,7 @@ cfg_if::cfg_if! {
                     "symlink_loop" => true,
                     "clock_time_get" => true,
                     "truncation_rights" => true,
+                    "path_filestat" => true,
```
kubkon: I guess that this PR should enable this test on Windows, right? :-)

Issue #16, opened 2019-05-14
Add `nix` support for missing host function calls.


sunfishcode:
At a quick glance, the following functions are not yet provided by [`nix`](https://github.com/nix-rust/nix/), but would be useful to have in `nix` for use by wasi-common: `mkdirat`, `unlinkat`, `renameat`, `linkat`, `clock_getres`, `clock_gettime`, `posix_fallocate`, `posix_fadvise`.

Currently code in wasi-common and lucet-wasi is calling libc interfaces for these functions directly, which requires unsafe code. Having `nix` provide these will mean the unsafe code is shifted to `nix` instead. This may not seem like a win on a theoretical level, but the more we can push `unsafe` code into dedicated utilities that do simple unsafe things with safe, testable, interfaces, the easier it is to audit everything.

I've marked this as "help wanted" -- this is not an urgent project, but it is a nicely scoped, straightforward, and subdividable project, and it's a good chance to see some of the low-level parts of the software stack that wasi-common is built on. And mentoring is available :-).

Additional resources:
 - Documentation for these functions in POSIX can be found here: http://pubs.opengroup.org/onlinepubs/9699919799/
 - [Here's an example](https://github.com/nix-rust/nix/pull/1050) of implementing a very simple function in `nix`, though the functions above are more complex and will need testcases :-).

jlb6740:
Hi .. I was hoping to be able to help on the wasi-common request and this might be a good place to start. I have an initial patch here:  [](https://github.com/nix-rust/nix/compare/master...jlb6740:implement_unlinkat?expand=1). I am new to rust so could use some pointers on testing this. Once this initial one is tested I'll submit a pull request and see about implementing some of the others.

sunfishcode:
Hi! Thanks for starting this! The patch looks good. For testing `unlinkat`, take a look at `test_fchmodat` in test/test_stat.rs. The basic setup it does is essentially what you want here: it creates a temp directory and a file inside that directory, obtains a directory file descriptor, and then does operations using that file descriptor plus a relative path. For your test, the operation should be to do `unlinkat` to remove the file. For completeness, you can also create a subdirectory inside the temp directory with `mkdir` and then remove it with `unlinkat` with `RemoveDir`.

Also, sorry, I accidentally hit the wrong button while looking at the link above and opened a PR. I've now closed it. We can open a new PR once the testcase is ready.

jlb6740:
[...]

PR #64, opened 2019-08-19
Implement path_filestat_set_times using nix instead of libc.

marmistrz:
As advised in #16.

Obsoletes #61. Still needs tests from https://github.com/CraneStation/wasi-misc-tests/pull/19. 

There are two calls to transmute left because `TimeSpec` from nix doesn't expose `UTIME_OMIT` and `UTIME_NOW`. This PR will stay as draft until support for those special `TimeSpec` values is added to nix.


kubkon:
LGTM! Even though it'd be nice to have `UTIME_OMIT` and `UTIME_NOW` exposed by `nix`, I'll merge it as-is for now as it nicely cleaned up that bit of code :-)

PR #65, opened 2019-08-19
Add the test for path_filestat_*

marmistrz:
From https://github.com/CraneStation/wasi-misc-tests/pull/19


marmistrz:
`path_filestat_set_times` seems broken. While `st_mtim` is a correct 2019-era timestamp, the underlying `utimensat` syscall is executed with `tv_sec = 1` (checked using `strace -f cargo test path_filestat`). This happens both when current master is being used or my refactor using nix from #64.

kubkon (approved):
LGTM!

PR #66, opened 2019-08-20
Fix filestat_from_nix conversion fn

kubkon:
Fixes bug uncovered in #65.

I think that in the long run we should consider utilising [filetime](https://github.com/alexcrichton/filetime), but this would require adding an extension trait so that we can set and get {a,c,m}times from `(&File, &Path)` pairs as well (essentially allowing for *at syscalls). @alexcrichton do you think it would be valid to extend `filetime` with such functionality for *nix hosts?

### DIFF
--- a/src/sys/unix/host_impl.rs
+++ b/src/sys/unix/host_impl.rs
@@ -201,20 +201,29 @@ pub(crate) fn filestat_from_nix(
 ) -> Result<host::__wasi_filestat_t> {
     use std::convert::TryFrom;
 
+    fn filestat_to_timestamp(secs: u64, nsecs: u64) -> Result<host::__wasi_timestamp_t> {
+        secs.checked_mul(1_000_000_000)
+            .and_then(|sec_nsec| sec_nsec.checked_add(nsecs))
+            .ok_or(host::__WASI_EOVERFLOW)
+    }
+
     let filetype = nix::sys::stat::SFlag::from_bits_truncate(filestat.st_mode);
     let dev =
         host::__wasi_device_t::try_from(filestat.st_dev).map_err(|_| host::__WASI_EOVERFLOW)?;
     let ino =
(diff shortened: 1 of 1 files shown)


kubkon:
cc @sunfishcode @marmistrz

alexcrichton:
Hm I'm not sure I fully understand, what would be added to `filetime`? (it's probably reasonable to add I'm just curious)

kubkon:
@alexcrichton Oh sorry for not explaining myself very well! Currently in `filetime` we can set/get times from handles such as `&File`. For WASI on *nix hosts, it'd be very useful to permit for `openat`-style calls, so then `&File` handle would point to a directory and `&Path` to a path relative to that handle. I'm just worried that representing open handle to a directory wrapped in `&File` would not be very clear from the interface perspective to the crate's user; that is, it could cause confusion where users would supply a handle to an actual file and not a directory, but perhaps the very same problem is still faced in raw syscalls to `openat`-style functions.

marmistrz (approved):
What a nasty bug! Thanks for catching that.

marmistrz:
> I'm just worried that representing open handle to a directory wrapped in &File would not be very clear from the interface perspective to the crate's user; that is, it could cause confusion where users would supply a handle to an actual file and not a directory, but perhaps the very same problem is still faced in raw syscalls to openat-style functions.

@kubkon it may cause confusion, but the `openat`-style functions are very confusing themselves. I didn't understand why in world one would want such syscalls until I read `openat(2)` :P

alexcrichton:
@kubkon sounds reasonable to me to add!

kubkon:
@marmistrz that’s a fair enough actually. I sometimes also find `openat`-style functions confusing :-)

@alexcrichton great! Let’s merge these changes as they are for now, with the intention of enabling the `openat`-style functionality in `filetime`, which I’ll do my best to add in the near future :-)

sunfishcode:
Sounds good!

PR #69, opened 2019-08-21
Update the pinned wasmtime revision.

sunfishcode:
This resynchronizes the versions of cranelift used in wasi-common
and wasmtime to 0.40.


kubkon (approved):
LGTM!

PR #71, opened 2019-08-22
Implement fd_readdir on Unix with nix & on Windows with libstd

marmistrz:
Before this can be merged:
* [x] we need to clarify and extend the test in https://github.com/CraneStation/wasi-misc-tests/pull/21. 
* [ ] On Unix, raw `readdir` executed on an empty directory will return two entries: `.` and `..`. On the other hand, `std::fs::read_dir` will return an empty iterator. Should we emulate the behavior of the unix `readdir` syscall in wasi by manually adding `.` and `..`? The current implementation does it but the spec is not clear about this. 
* [ ] we're leaking inode/serial number information about `..`, which may not be preopened and may hence be inaccessible for apps run inside WASI. I don't think it's an issue, but I want to make sure this isn't a vulnerability
* [ ] upstream the changes done for nix (tracking in https://github.com/nix-rust/nix/issues/1110)
* [ ] mac is currently broken
-----
Follow-ups needed:
* Add a heuristic optimization to `fd_readdir` on Windows: the most common use case where seeking may be needed is when the buffer length is insufficient. We could cache the last cookie and the relevant handle for each file descriptor so that the implementation is O(n) and not O(n^2) with O(1) buffer in the most common case
* Add `Into`/`From` conversions for `Filetype` 
* Raise the issues about the spec  of `fd_readdir`:
  * `readdir(3p)` says that 
     > If a file is removed from or added to the directory after the most recent call to  opendir()  or
     > rewinddir(),  whether  a subsequent call to readdir() returns an entry for that file is unspecified.

    We need to add this to spec of `fd_readdir` in some form.
  * discuss how to clarify the semantics of cookie (also that it corresponds to the next entry)
  * mention the similarity to `readdir`/`seekdir` in the docs
  * should the implementation return  `.` & `..`?
  * be precise whether or not the filenames are NULL-terminated (in POSIX they are)
* ~nix uses `readdir_r` which is deprecated in `readdir_r(3)`. Raise an issue and discuss whether or not `readdir` should be used instead (at least on Linux)~ https://github.com/rust-lang/rust/issues/34668
-------
**Implementation notes:**
The code written using libstd could be mostly reused for Linux, but:
* (first and foremost) this would force us to emulate entries for `.` & `..`, which would cost us extra syscalls on Linux
* accessors for `file_serial_no` not stable yet: https://github.com/rust-lang/rust/issues/63010
* `ReadDir` from `libstd` requires to have a `Path` for no real reason.

While we could technically either hack the `std::fs::ReadDir` using unsafe code (transmute) or fork it, I believe that the overhead of adding at least two (and probably at least four) extra syscalls 

`ReadDir` can only be constructed from a root path for the directory query.

On Unix, the path is only used for `impl Debug` and `fn path`. In other words, `fd_readdir` could be implemented completely using `std::fs::ReadDir` if we just removed every reference to the root path in the code and created it from a `&File` instead. Unfortunately, this wouldn't be a backwards compatible change in libstd, because `fn path` would have to return an `Option<Path>`.

On Windows, while the root path is used in `impl Iterator for ReadDir`, it's only seeming that the restriction is required, since one can recover path from `&File` on Windows. 

Therefore, to all my knowledge it appears that the explicit path could be removed from `ReadDir`, but it would break backwards compatibility, so such a change won't unfortunately be accepted into stdlib.

-------
What are the alternatives?
* we could manually create our own `ReadDir` and then use `std::mem::transmute` to convert it to `std::fs::ReadDir`. While this will work, it is heavily dependent on non-public Rust libstd API and its implementation details. We'd also probably have to do our own wrapper to disallow `DirEntry::path`
* do a copy-paste of `ReadDir` (possibly in a separate crate), dropping the requirement on `Path`. This has the obvious disadvantage of maintaining our own fork but allows us to have a common implementation for Unix & Windows. We can't actually reuse the `readdir` syscall on Unix because of implementation details: on Unix the filenames are null-terminated, on wasi they aren't (cf. the discussion in https://github.com/CraneStation/wasi-misc-tests/pull/21)
* implement `fd_readdir` using `nix` on unix, `ReadDir` on windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 Cargo.lock
 **/*.rs.bk
 .DS_Store
+.vscode
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ log = "0.4"
 filetime = "0.2.7"
 
 [target.'cfg(unix)'.dependencies]
-nix = "0.15"
+nix = { git = "https://github.com/marmistrz/nix", branch = "seek" }
 
 [target.'cfg(windows)'.dependencies]
 winx = { path = "winx" }
--- a/src/hostcalls_impl/fs.rs
+++ b/src/hostcalls_impl/fs.rs
@@ -8,7 +8,7 @@ use crate::sys::hostcalls_impl::fs_helpers::path_open_rights;
 use crate::sys::{errno_from_ioerror, host_impl, hostcalls_impl};
 use crate::{host, wasm32, Result};
 use filetime::{set_file_handle_times, FileTime};
-use log::trace;
+use log::{debug, trace};
 use std::convert::TryInto;
 use std::fs::File;
 use std::io::{self, Read, Seek, SeekFrom, Write};
@@ -626,13 +626,14 @@ pub(crate) fn fd_readdir(
 
     let mut host_buf = dec_slice_of_mut::<u8>(memory, buf, buf_len)?;
 
-    trace!("     | (buf,buf_len)={:?}", host_buf);
+    trace!("     | before: (buf,buf_len)={:?}", host_buf);
 
     let cookie = dec_dircookie(cookie);
 
-    let iter = hostcalls_impl::fd_readdir(fd, host_buf, cookie)?;
+    let vec = hostcalls_impl::fd_readdir(fd, cookie)?;
+    debug!("Received dirents: {:?}", vec);
     let mut used = 0;
-    for dirent in iter {
+    for dirent in vec {
         let dirent_raw = dirent.to_raw()?;
         let offset = dirent_raw.len();
         host_buf[0..offset].copy_from_slice(&dirent_raw);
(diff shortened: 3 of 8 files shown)


kubkon:
@marmistrz have a look at WebAssembly/WASI#61 at some discussion about `fd_readdir` impl on Windows hosts.

marmistrz:
It looks like upstreaming the changes to nix is a non-trivial task. I think the best idea is to currently "fork" the readdir module from nix and make the changes locally in our source tree (i.e. add ` src/sys/unix/hostcalls_impl/dir.rs`). Are you okay with that?

sunfishcode:
To make sure I'm not missing something, is this PR subsumed by #116?

kubkon:
> To make sure I'm not missing something, is this PR subsumed by #116?

That's the idea, yep.

kubkon:
It actually might make sense to close it manually now so as to avoid any future confusion.

marmistrz:
Yes, no further work will happen on this branch.

PR #72, opened 2019-08-22
Use safe casts in path_filestat_set_times on unix.

marmistrz:
Unsafe casts were introduced in #64.


kubkon:
I think you meant `path_filestat_set_times` rather than `fd_readdir` :-)

marmistrz:
Yeah :) fixed

kubkon (approved):
LGTM, thanks!

PR #74, opened 2019-08-23
Fixes #30

### LINKED ISSUE #30: Use safe code for calling `readlinkat`
On Linux, `readlinkat` returns `EINVAL` if given a zero-length buffer. WASI's `wasi_path_readlink` should follow the POSIX semantics for `readlinkat` instead. wasi-common implements this, however it uses an `unsafe` call to the libc `readlinkat` rather than calling through a safe nix wrapper.

See the discussion [here](https://github.com/CraneStation/wasi-common/pull/27/#discussion_r294497606) for more details.

kubkon:
Currently, the fixed version of `readlinkat` is provided in upstream
[nix](https://github.com/nix-rust/nix) crate only, therefore, we
need to change `nix` to be a git dependency, which is not ideal.
However, a new release should be out soon-ish, hence, I reckon it's
OK to go with it as-is for now, and change to a semver when a new
version gets released.

Changes:
* uses `nix::fcntl::readlinkat` instead of a direct call to
  `libc::readlinkat`
* if read target path is longer than the buffer provided,
  pads the buffer with 0s

Kudos to @sendilkumarn for providing a nice fix to `nix` (nix-rust/nix#1109).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ log = "0.4"
 filetime = "0.2.7"
 
 [target.'cfg(unix)'.dependencies]
-nix = "0.15"
+nix = { git = "https://github.com/nix-rust/nix" }
 
 [target.'cfg(windows)'.dependencies]
 winx = { path = "winx" }
--- a/src/sys/unix/hostcalls_impl/fs.rs
+++ b/src/sys/unix/hostcalls_impl/fs.rs
@@ -274,34 +274,20 @@ pub(crate) fn fd_readdir(
 }
 
 pub(crate) fn path_readlink(resolved: PathGet, buf: &mut [u8]) -> Result<usize> {
-    use nix::errno::Errno;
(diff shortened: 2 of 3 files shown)


kubkon:
cc @alexcrichton @alexcrichton

marmistrz:
readlinkat should be supported on 0.15.0: https://docs.rs/nix/0.15.0/nix/fcntl/fn.readlinkat.html

marmistrz (approved):
Please use a more meaningful commit message. Otherwise, LGTM

sendilkumarn (approved):
Looks good to me 👍

kubkon:
> Please use a more meaningful commit message. Otherwise, LGTM

Oh sorry, actually the commit message has got a very short summary indeed but a very verbose description if you unfold it :-) Nonetheless, in the future, I'll try and make the summaries slightly more verbose.

kubkon:
Due to the fact we've got a couple of versions of `wasi-common` in [crates.io](https://crates.io), I'd like to wait with merging this change into upstream until the applied `nix` fix is behind a semver so that we don't rely on any `git` deps.

PR #75, opened 2019-08-23
Move fd_advise on Linux to nix 

marmistrz:
Bases on #74.
We could actually enable `posix_fadvise` for all platforms supported by nix, but it's going to be messy. https://github.com/nix-rust/nix/blob/master/src/fcntl.rs#L15


kubkon:
> Bases on #74.
> We could actually enable `posix_fadvise` for all platforms supported by nix, but it's going to be messy. https://github.com/nix-rust/nix/blob/master/src/fcntl.rs#L15

Nice, looks good, thanks! As far as enabling for more platforms, sure, but I'd leave that for the (near) future. I'd rather we finished basic (experimental) support for Windows first, before diverging our focus on other things :-)

kubkon (approved):

PR #77, opened 2019-08-25
[WIP] Add missing docs to functions, constants and modules

kubkon:
This PR adds missing docs to functions, constants and modules.

TODO:
- [ ] `hostcalls` module
- [ ] `memory` module
- [ ] constants
- [ ] `WasiCtx` and related

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -1,6 +1,4 @@
-//! WASI host types as defined in host. This file was originally generated
-//! by running bindgen over wasi/core.h, and the content
-//! still largely reflects that.
+//! WASI host types as defined in host.
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 use crate::Result;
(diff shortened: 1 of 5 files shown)


sunfishcode:
My thought here is it's not worth spending a lot of time documenting the hostcalls, `memory`, and constants right now, as it's possible we'll be able to auto-generate a lot of those from the new witx files.

So my suggestion here is to focus on documenting `WasiCtx` for now :-).

kubkon:
> My thought here is it's not worth spending a lot of time documenting the hostcalls, `memory`, and constants right now, as it's possible we'll be able to auto-generate a lot of those from the new witx files.
> 
> So my suggestion here is to focus on documenting `WasiCtx` for now :-).

Sounds good! Should I close this PR then, or would you like to me retarget it with `WasiCtx` docs?

sunfishcode:
If you still want to land the changes in src/host.rs, src/macros.rs, etc., we could still do that. But for new changes, I generally prefer a new PR over purposing an existing one.

kubkon:
> If you still want to land the changes in src/host.rs, src/macros.rs, etc., we could still do that. But for new changes, I generally prefer a new PR over purposing an existing one.

It's OK. I think I'd prefer to leave it untouched for now and see how we get on with auto-docs from the witx files :-) Closing this for now.

PR #78, opened 2019-08-26
Fix incorrect fd_readdir behaviour

kubkon:
* ~~when executed twice in a row, need to manually reset the stream
  by calling seekdir with `__WASI_DIRCOOKIE_START`, if `__WASI_DIRCOOKIE_START` was specified~~
I've gone ahead and rewritten majority of `fd_readdir` using `nix` crate as I couldn't get `seekdir` to work properly on a macos when trying to pass the tests in #76 
* fix mapping between d_type and `__wasi_filetype_t`
* include minor refactor - removes use of `wasm32` module on the host's side

cc @marmistrz

### DIFF
--- a/src/sys/unix/host_impl.rs
+++ b/src/sys/unix/host_impl.rs
@@ -2,7 +2,7 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 #![allow(dead_code)]
-use crate::{host, memory, wasm32, Result};
+use crate::{host, memory, Result};
 use std::ffi::OsStr;
 use std::os::unix::prelude::OsStrExt;
 
@@ -228,29 +228,54 @@ pub(crate) fn filestat_from_nix(
     })
 }
 
+pub(crate) fn dirent_filetype_from_host(
+    host_entry: &nix::libc::dirent,
+) -> Result<host::__wasi_filetype_t> {
+    match host_entry.d_type {
+        libc::DT_FIFO => Ok(host::__WASI_FILETYPE_UNKNOWN),
+        libc::DT_CHR => Ok(host::__WASI_FILETYPE_CHARACTER_DEVICE),
+        libc::DT_DIR => Ok(host::__WASI_FILETYPE_DIRECTORY),
+        libc::DT_BLK => Ok(host::__WASI_FILETYPE_BLOCK_DEVICE),
+        libc::DT_REG => Ok(host::__WASI_FILETYPE_REGULAR_FILE),
+        libc::DT_LNK => Ok(host::__WASI_FILETYPE_SYMBOLIC_LINK),
+        libc::DT_SOCK => {
+            // TODO how to discriminate between STREAM and DGRAM?
+            // Perhaps, we should create a more general WASI filetype
+            // such as __WASI_FILETYPE_SOCKET, and then it would be
+            // up to the client to check whether it's actually
+            // STREAM or DGRAM?
+            Ok(host::__WASI_FILETYPE_UNKNOWN)
+        }
+        libc::DT_UNKNOWN => Ok(host::__WASI_FILETYPE_UNKNOWN),
+        _ => Err(host::__WASI_EINVAL),
+    }
+}
+
 #[cfg(target_os = "linux")]
-pub(crate) fn dirent_from_host(host_entry: &nix::libc::dirent) -> Result<wasm32::__wasi_dirent_t> {
-    let mut entry = unsafe { std::mem::zeroed::<wasm32::__wasi_dirent_t>() };
+pub(crate) fn dirent_from_host(host_entry: &nix::libc::dirent) -> Result<host::__wasi_dirent_t> {
+    let mut entry = unsafe { std::mem::zeroed::<host::__wasi_dirent_t>() };
     let d_namlen = unsafe { std::ffi::CStr::from_ptr(host_entry.d_name.as_ptr()) }
         .to_bytes()
         .len();
     if d_namlen > u32::max_value() as usize {
         return Err(host::__WASI_EIO);
     }
+    let d_type = dirent_filetype_from_host(host_entry)?;
     entry.d_ino = memory::enc_inode(host_entry.d_ino);
     entry.d_next = memory::enc_dircookie(host_entry.d_off as u64);
     entry.d_namlen = memory::enc_u32(d_namlen as u32);
-    entry.d_type = memory::enc_filetype(host_entry.d_type);
+    entry.d_type = memory::enc_filetype(d_type);
     Ok(entry)
 }
 
 #[cfg(not(target_os = "linux"))]
-pub(crate) fn dirent_from_host(host_entry: &nix::libc::dirent) -> Result<wasm32::__wasi_dirent_t> {
-    let mut entry = unsafe { std::mem::zeroed::<wasm32::__wasi_dirent_t>() };
+pub(crate) fn dirent_from_host(host_entry: &nix::libc::dirent) -> Result<host::__wasi_dirent_t> {
+    let mut entry = unsafe { std::mem::zeroed::<host::__wasi_dirent_t>() };
+    let d_type = dirent_filetype_from_host(host_entry)?;
     entry.d_ino = memory::enc_inode(host_entry.d_ino);
     entry.d_next = memory::enc_dircookie(host_entry.d_seekoff);
     entry.d_namlen = memory::enc_u32(u32::from(host_entry.d_namlen));
-    entry.d_type = memory::enc_filetype(host_entry.d_type);
+    entry.d_type = memory::enc_filetype(d_type);
     Ok(entry)
 }
 
--- a/src/sys/unix/hostcalls_impl/fs.rs
+++ b/src/sys/unix/hostcalls_impl/fs.rs
@@ -5,7 +5,7 @@ use crate::helpers::systemtime_to_timestamp;
 use crate::hostcalls_impl::PathGet;
 use crate::sys::errno_from_ioerror;
 use crate::sys::host_impl;
-use crate::{host, wasm32, Result};
+use crate::{host, Result};
 use nix::libc::{self, c_long, c_void};
 use std::convert::TryInto;
 use std::ffi::CString;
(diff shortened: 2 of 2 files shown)


src/sys/unix/host_impl.rs
```diff
@@ -228,29 +228,47 @@ pub(crate) fn filestat_from_nix(
...
+    host_entry: &nix::libc::dirent,
+) -> Result<host::__wasi_filetype_t> {
+    match host_entry.d_type {
+        libc::DT_FIFO => Ok(host::__WASI_FILETYPE_SOCKET_STREAM),
+        libc::DT_CHR => Ok(host::__WASI_FILETYPE_CHARACTER_DEVICE),
+        libc::DT_DIR => Ok(host::__WASI_FILETYPE_DIRECTORY),
+        libc::DT_BLK => Ok(host::__WASI_FILETYPE_BLOCK_DEVICE),
+        libc::DT_REG => Ok(host::__WASI_FILETYPE_REGULAR_FILE),
+        libc::DT_LNK => Ok(host::__WASI_FILETYPE_SYMBOLIC_LINK),
+        libc::DT_SOCK => Ok(host::__WASI_FILETYPE_SOCKET_DGRAM),
```
marmistrz: This is incorrect. 
```
              DT_SOCK     This is a UNIX domain socket.
```
A Unix domain socket can be [`SOCK_DGRAM` or `SOCK_STREAM` or `SOCK_SEQPACKET`](https://en.wikipedia.org/wiki/Unix_domain_socket).

Please at least add a `TODO` comment similar to the one in `fs.rs`, `fn metadata`:
```
        // TODO we should use getsockopt to find out if it's
        // SOCKET_STREAM or SOCKET_DGRAM
```

kubkon:
cc @jedisct1 @alexcrichton

src/sys/unix/hostcalls_impl/fs.rs
```diff
@@ -221,9 +221,7 @@ pub(crate) fn fd_readdir(
     if dir.is_null() {
         return Err(host_impl::errno_from_nix(nix::errno::Errno::last()));
     }
-    if cookie != wasm32::__WASI_DIRCOOKIE_START {
-        unsafe { seekdir(dir, cookie as c_long) };
-    }
+    unsafe { seekdir(dir, cookie as c_long) };
```
marmistrz: Please add a comment why this is needed. At first sight it appears that since we're doing new `fdopendir` for every call, this should not apparently be needed.

marmistrz (changes requested):

src/sys/unix/host_impl.rs
kubkon: This is not that simple. For `getsockopt` call we need an fd to run the check on, and we don't have it in `fd_readdir` call. Perhaps, we should create a more general WASI filetype such as `__WASI_FILETYPE_SOCKET`, and then it would be up to the client to check whether it's actually `STREAM` or `DGRAM`? Anyhow, I've added a comment to that effect now.

src/sys/unix/hostcalls_impl/fs.rs
```diff
@@ -213,52 +213,66 @@ pub(crate) fn fd_readdir(
...
     let mut host_buf_offset: usize = 0;
-    while left > 0 {
-        let mut host_entry: *mut dirent = std::ptr::null_mut();
-        let res = unsafe { readdir_r(dir, &mut entry_buf, &mut host_entry) };
-        if res == -1 {
-            return Err(host_impl::errno_from_nix(nix::errno::Errno::last()));
+
+    for (count, host_entry) in dir.iter().enumerate() {
+        let count: u64 = count.try_into().map_err(|_| host::__WASI_EOVERFLOW)?;
+        if count < cookie {
```
marmistrz: Why do you assume that `cookie` is monotone? The manpage on Linux explicitly says that:

> Applications should treat this field as an opaque value, making no  assumptions  about
              its contents

kubkon: That’s a fair point. It was a quick fix just to see if it works on macos. It turns out `seekdir` doesn’t iterate the dir stream on Mac properly for some reason. Anyhow, I’ll provide a fix and report back.

kubkon: I've provided a fix, have a look and let me know what you reckon.

src/sys/unix/host_impl.rs
sunfishcode: Agreed; I don't think it's possible to determine whether a unix-domain socket is stream, dgram, or something else without opening it.

FWIW, CloudABI handles this by [just returning `FILETYPE_SOCKET_STREAM`](https://github.com/NuxiNL/cloudabi-utils/blob/master/src/libemulator/posix.c#L1689).

What if we returned `__WASI_FILETYPE_UNKNOWN` for `DT_SOCK` for now? That ought to suffice in many cases, because it isn't generally possible to interact with a Unix-domain socket without knowing a fair amount about who you'll be talking to on the other end, and if you know that, you typically don't need `readdir` to tell you that the thing you're opening is a socket.

src/sys/unix/host_impl.rs
```diff
@@ -228,29 +228,54 @@ pub(crate) fn filestat_from_nix(
     })
 }
 
+pub(crate) fn dirent_filetype_from_host(
+    host_entry: &nix::libc::dirent,
+) -> Result<host::__wasi_filetype_t> {
+    match host_entry.d_type {
+        libc::DT_FIFO => Ok(host::__WASI_FILETYPE_SOCKET_STREAM),
```
sunfishcode: Same applies to `DT_FIFO` -- `FILETYPE_SOCKET_STREAM` isn't entirely accurate here, so what if we used `FILETYPE_UNKNOWN` for now?

src/sys/unix/hostcalls_impl/fs.rs
```diff
@@ -213,52 +213,68 @@ pub(crate) fn fd_readdir(
...
+            // For some reason, on MacOS, dirent::d_seekoff is never incremented, so
+            // we need to emulate by filling in the blanks ourselves.
+            let count: host::__wasi_dircookie_t =
+                count.try_into().map_err(|_| host::__WASI_EOVERFLOW)?;
+            entry.d_next = count + 1;
         }
-        if host_entry.is_null() {
-            break;
+
+        if entry.d_next <= cookie {
```
sunfishcode: If I read this right, this means that a program which reads a large directory with a small buffer will end up doing a quadradic amount of work, because each call will have to iterate forward from the beginning of the directory. If there's really no other option on MacOS, but I'd like to dig into this a little more to understand what the alternatives are.

src/sys/unix/hostcalls_impl/fs.rs
```diff
@@ -213,52 +213,68 @@ pub(crate) fn fd_readdir(
...
 
     let host_buf_ptr = host_buf.as_mut_ptr();
     let host_buf_len = host_buf.len();
-    let dir = unsafe { fdopendir(fd.as_raw_fd()) };
-    if dir.is_null() {
-        return Err(host_impl::errno_from_nix(nix::errno::Errno::last()));
-    }
-    if cookie != wasm32::__WASI_DIRCOOKIE_START {
-        unsafe { seekdir(dir, cookie as c_long) };
-    }
```
sunfishcode, line 226: While here, I noticed that we're not currently doing `rewinddir` to rewind the directory in the case of `__WASI_DIRCOOKIE_START`.

Also, with the current code not using seekdir but instead just seeking until `d_next` passes the cookie value, it seems like we should do a rewind before doing that.

src/sys/unix/host_impl.rs
marmistrz: This also applies to `filetype_from_nix` (where this conversion originally comes from) and `filetype` (which was based on the former) in `src/sys/unix/hostcalls_impl/fs.rs`.

I guess we should change all of them in a separate PR and/or discuss whether or not the WASI standard should enforce a particular mapping of fifos.

marmistrz:
I'm not sure if this change makes sense anymore, it's getting a patchwork of nix and non-nix. I'm going to rewrite `fd_readdir` entirely using nix in the coming days (hopefully tomorrow),  reusing as much code between windows and unix.

I suggest we close this change and just throw away the old implementation in a subsequent PR, possibly leaving an issue which will be closed by the #71 once it's augmented with a nix-based implementation for unix.

src/sys/unix/hostcalls_impl/fs.rs
kubkon: That’s a good observation. So `nix` internally does a `rewinddir` when the `Dir` iterator gets dropped.

src/sys/unix/hostcalls_impl/fs.rs
kubkon: That’s very true, thanks for pointing that out. MacOS got me stumbled here. Originally I didn’t want to redo the implementation of `fd_readdir`—it was more of a check to see if it can pass all tests at all, and it did but using this approach; and from what I can see, using `nix`’s `Dir` iterative will always imply doing quadratic work.

kubkon:
> I'm not sure if this change makes sense anymore, it's getting a patchwork of nix and non-nix. I'm going to rewrite `fd_readdir` entirely using nix in the coming days (hopefully tomorrow), reusing as much code between windows and unix.
> 
> I suggest we close this change and just throw away the old implementation in a subsequent PR, possibly leaving an issue which will be closed by the #71 once it's augmented with a nix-based implementation for unix.

I have to disagree on that one at this time, mainly due to the two facts that @sunfishcode pointed out: 1) with `nix` it seems like doing quadratic work, and 2) it should be possible to get `fd_readdir` to work using POSIX syscalls on MacOS. I suggest we revert the latest commit to this PR, and I'll investigate closer what is happening on MacOS. In the meantime, you could suggest an alternative approach as well targeting mainly Windows and how we should deal with `wasi_dircookie_t` there. How does that sound for you guys, @sunfishcode @marmistrz?

src/sys/unix/host_impl.rs
kubkon: @sunfishcode sounds fair! I'll adjust the return types for `dirent` structs in this PR already. @marmistrz feel free to open another PR to adjust them elsewhere as well.

sunfishcode (approved):
Landing the bug-fix parts of this PR first and figuring out `nix` or other issues separately sounds good to me!

kubkon:
OK then, I'll merge it as-is for now, and create another PR with a proposed fix for macOS and BSD-like *nixes and we'll take it from there I guess.

PR #79, opened 2019-08-26
Check for socket type in filetype.

marmistrz:
The signature for Windows has also been changed, so that the function
can be reused in cross-platform modules. 

Unfortunately, this makes `fd_filestat_get` require an extra syscall on Unix, because the data returned by `fstat` doesn't include the information about the socket type.

cc @kubkon

### DIFF
--- a/src/sys/windows/hostcalls_impl/fs.rs
+++ b/src/sys/windows/hostcalls_impl/fs.rs
@@ -288,11 +288,11 @@ pub(crate) fn fd_filestat_get_impl(file: &std::fs::File) -> Result<host::__wasi_
             .modified()
             .map_err(errno_from_ioerror)
             .and_then(systemtime_to_timestamp)?,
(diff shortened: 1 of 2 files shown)


src/sys/windows/hostcalls_impl/fs.rs
```diff
@@ -288,11 +288,11 @@ pub(crate) fn fd_filestat_get_impl(file: &std::fs::File) -> Result<host::__wasi_
             .modified()
             .map_err(errno_from_ioerror)
             .and_then(systemtime_to_timestamp)?,
-        st_filetype: filetype(&metadata).map_err(errno_from_ioerror)?,
+        st_filetype: filetype(&metadata)?,
```
kubkon:
```suggestion
        st_filetype: filetype(file, &metadata)?,
```

kubkon:
LGTM, thanks!

PR #80, opened 2019-08-27
Treat FIFOs' type as __WASI_FILETYPE_UNKNOWN. 

marmistrz:
This came up in #78.
cc @kubkon

### DIFF
--- a/src/sys/unix/fdentry_impl.rs
+++ b/src/sys/unix/fdentry_impl.rs
@@ -51,10 +51,7 @@ pub(crate) fn determine_type_rights<Fd: AsRawFd>(
         // we just make a `File` here for convenience; we don't want it to close when it drops
         let file =
             std::mem::ManuallyDrop::new(unsafe { std::fs::File::from_raw_fd(fd.as_raw_fd()) });
-        let ft = file
-            .metadata()
-            .map_err(errno_from_ioerror)?
-            .file_type();
+        let ft = file.metadata().map_err(errno_from_ioerror)?.file_type();
         if ft.is_block_device() {
             (
                 host::__WASI_FILETYPE_BLOCK_DEVICE,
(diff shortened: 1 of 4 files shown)


src/sys/unix/fdentry_impl.rs
```diff
@@ -110,9 +107,9 @@ pub(crate) fn determine_type_rights<Fd: AsRawFd>(
             }
         } else if ft.is_fifo() {
             (
-                host::__WASI_FILETYPE_SOCKET_STREAM,
-                host::RIGHTS_SOCKET_BASE,
-                host::RIGHTS_SOCKET_INHERITING,
+                host::__WASI_FILETYPE_UNKNOWN,
+                host::RIGHTS_REGULAR_FILE_BASE,
+                host::RIGHTS_REGULAR_FILE_INHERITING,
```
sunfishcode, line 112: Here's a subtle question. What should the default rights be for a fifo? Fifos don't support many of the regular file operations like `fd_seek` and so on. Should we add a new `host::RIGHTS_FIFO_BASE` which looks like `RIGHTS_SOCKET_BASE` but without `SOCK_SHUTDOWN`, or should we let it have all the file rights and just let them fail with EBADF if anyone tries to use them?

marmistrz: WASI currently doesn't really support FIFOs, so I think we can currently just treat them as regular files and let them fail with `EBADF`. I'm wondering how some Windows ext4 drivers treat FIFOs. (I only know such drivers exist)

Alternatively, there's some rationale for leaving the rights as they currently are, because FIFOs are to some extent similar to sockets.

As far as I know, there's standardization of sockets pending, so this will probably be a good opportunity to also discuss how to deal with FIFOs.

sunfishcode: Ok. I filed https://github.com/WebAssembly/WASI/issues/85 to track this.

sunfishcode (approved):

PR #81, opened 2019-08-28
Fix fd_readdir on BSD-style nixes

kubkon:
The fix was tested on Darwin-XNU and FreeBSD. The change introduces thread-safe cache of `(RawFd, *mut libc::DIR)` pairs so that `libc::fdopendir` syscall is called only once when invoking `fd_readdir` for the first time, and then the pointer to the directory stream, `*mut libc::DIR`, is reused until the matching raw file descriptor is closed.

This fix allows then correct use (and matching to the implementation on Linux kernels) of `libc::seekdir` and `libc::rewinddir` to seek through and rewind the existing directory stream, `*mut libc::DIR`, which otherwise seems to be reset/invalidated every time `libc::fdopendir` is called (unlike on Linux, where this behaviour is not observed).

@sunfishcode @marmistrz to sum up, on BSD-based kernels, we can either go with something similar to the proposed solution in this PR, or we can traverse the entire directory stream from the beginning at every `fd_readdir` call thus doing quadratic work. Let me know your thoughts!

To provide some more context, the following snippet does not behave as expected on BSD-based kernels (although this particular code applies only to Darwin-XNU):

```rust
use std::ffi::CString;
use std::ptr;

fn main() {
    unsafe {
        let name = CString::new(".").unwrap();
        let dir = libc::opendir(name.as_ptr());
        libc::seekdir(dir, 2); // <- no effect on `dir` whatsoever; the "2" here is just an example

        loop {
            println!("loc={:?}", libc::telldir(dir));
            let entry = libc::readdir(dir);
            if entry == ptr::null_mut() {
                println!("errno={}", nix::errno::Errno::last());
                break;
            }
            println!("entry={:?}", *entry);
        }

        libc::closedir(dir);
    }
}
```

Whereas the following snippet successfully employs `seekdir`:

```rust
use std::ffi::CString;
use std::ptr;

fn main() {
    unsafe {
        let name = CString::new(".").unwrap();
        let dir = libc::opendir(name.as_ptr());

        loop {
            println!("loc={:?}", libc::telldir(dir));
            let entry = libc::readdir(dir);
            if entry == ptr::null_mut() {
                break;
            }
            println!("entry={:?}", *entry);
        }

        libc::seekdir(dir, 2); // <- this `seekdir` call is successful

        loop {
            println!("loc={:?}", libc::telldir(dir));
            let entry = libc::readdir(dir);
            if entry == ptr::null_mut() {
                break;
            }
            println!("entry={:?}", *entry);
        }

        libc::closedir(dir);
    }
}

```

### DIFF
--- a/src/fdentry.rs
+++ b/src/fdentry.rs
@@ -1,29 +1,82 @@
 use crate::sys::fdentry_impl;
 use crate::{host, Error, Result};
-
+use cfg_if::cfg_if;
 use std::mem::ManuallyDrop;
 use std::path::PathBuf;
 use std::{fs, io};
 
-#[derive(Debug)]
-pub enum Descriptor {
-    File(fs::File),
-    Stdin,
-    Stdout,
-    Stderr,
+cfg_if! {
+    if #[cfg(any(
+        target_os = "macos",
+        target_os = "freebsd",
+        target_os = "netbsd",
+        target_os = "openbsd",
+        target_os = "ios",
+        target_os = "dragonfly"
+    ))] {
+        use std::sync::Mutex;
+
+        #[derive(Debug)]
+        pub struct DirStream {
+            pub file: ManuallyDrop<fs::File>,
+            pub dir_ptr: *mut libc::DIR,
+        }
+
+        impl Drop for DirStream {
+            fn drop(&mut self) {
+                unsafe { libc::closedir(self.dir_ptr) };
+            }
+        }
+
+        #[derive(Debug)]
+        pub enum Descriptor {
+            File {
+                file: fs::File,
+                dir_stream: Option<Mutex<DirStream>>,
+            },
+            Stdin,
+            Stdout,
+            Stderr,
+        }
+
+        impl From<fs::File> for Descriptor {
+            fn from(file: fs::File) -> Self {
+                Descriptor::File {
+                    file,
+                    dir_stream: None
+                }
+            }
+        }
+    } else {
+        #[derive(Debug)]
+        pub enum Descriptor {
+            File {
+                file: fs::File,
+            },
+            Stdin,
+            Stdout,
+            Stderr,
+        }
+
+        impl From<fs::File> for Descriptor {
+            fn from(file: fs::File) -> Self {
+                Descriptor::File { file }
+            }
+        }
+    }
 }
 
 impl Descriptor {
     pub fn as_file(&self) -> Result<&fs::File> {
         match self {
-            Descriptor::File(f) => Ok(f),
+            Descriptor::File { file, .. } => Ok(file),
             _ => Err(Error::EBADF),
         }
     }
 
     pub fn is_file(&self) -> bool {
         match self {
-            Descriptor::File(_) => true,
+            Descriptor::File { .. } => true,
             _ => false,
         }
     }
@@ -80,7 +133,7 @@ impl FdEntry {
             |(file_type, rights_base, rights_inheriting)| Self {
                 fd_object: FdObject {
                     file_type,
-                    descriptor: ManuallyDrop::new(Descriptor::File(file)),
+                    descriptor: ManuallyDrop::new(Descriptor::from(file)),
                     needs_close: true,
                 },
                 rights_base,
--- a/src/fdentry.rs
+++ b/src/fdentry.rs
@@ -5,7 +5,7 @@ use std::path::PathBuf;
 use std::{fs, io};
 
 #[derive(Debug)]
-pub enum Descriptor {
+pub(crate) enum Descriptor {
     OsFile(OsFile),
     Stdin,
     Stdout,
@@ -13,21 +13,21 @@ pub enum Descriptor {
 }
 
 impl Descriptor {
-    pub fn as_file(&self) -> Result<&OsFile> {
+    pub(crate) fn as_file(&self) -> Result<&OsFile> {
         match self {
             Descriptor::OsFile(file) => Ok(file),
             _ => Err(Error::EBADF),
         }
     }
 
-    pub fn as_file_mut(&mut self) -> Result<&mut OsFile> {
+    pub(crate) fn as_file_mut(&mut self) -> Result<&mut OsFile> {
         match self {
             Descriptor::OsFile(file) => Ok(file),
             _ => Err(Error::EBADF),
         }
     }
 
-    pub fn is_file(&self) -> bool {
+    pub(crate) fn is_file(&self) -> bool {
         match self {
             Descriptor::OsFile(_) => true,
             _ => false,
@@ -35,7 +35,7 @@ impl Descriptor {
     }
 
     #[allow(unused)]
-    pub fn is_stdin(&self) -> bool {
+    pub(crate) fn is_stdin(&self) -> bool {
         match self {
             Descriptor::Stdin => true,
             _ => false,
@@ -43,7 +43,7 @@ impl Descriptor {
     }
 
     #[allow(unused)]
-    pub fn is_stdout(&self) -> bool {
+    pub(crate) fn is_stdout(&self) -> bool {
         match self {
             Descriptor::Stdout => true,
             _ => false,
@@ -51,7 +51,7 @@ impl Descriptor {
     }
 
     #[allow(unused)]
-    pub fn is_stderr(&self) -> bool {
+    pub(crate) fn is_stderr(&self) -> bool {
         match self {
             Descriptor::Stderr => true,
             _ => false,
@@ -60,19 +60,19 @@ impl Descriptor {
 }
 
 #[derive(Debug)]
-pub struct FdObject {
-    pub file_type: host::__wasi_filetype_t,
-    pub descriptor: ManuallyDrop<Descriptor>,
-    pub needs_close: bool,
+pub(crate) struct FdObject {
+    pub(crate) file_type: host::__wasi_filetype_t,
+    pub(crate) descriptor: ManuallyDrop<Descriptor>,
+    pub(crate) needs_close: bool,
     // TODO: directories
 }
 
 #[derive(Debug)]
-pub struct FdEntry {
-    pub fd_object: FdObject,
-    pub rights_base: host::__wasi_rights_t,
-    pub rights_inheriting: host::__wasi_rights_t,
-    pub preopen_path: Option<PathBuf>,
+pub(crate) struct FdEntry {
+    pub(crate) fd_object: FdObject,
+    pub(crate) rights_base: host::__wasi_rights_t,
+    pub(crate) rights_inheriting: host::__wasi_rights_t,
+    pub(crate) preopen_path: Option<PathBuf>,
 }
 
 impl Drop for FdObject {
@@ -84,7 +84,7 @@ impl Drop for FdObject {
 }
 
 impl FdEntry {
-    pub fn from(file: fs::File) -> Result<Self> {
+    pub(crate) fn from(file: fs::File) -> Result<Self> {
         determine_type_and_access_rights(&file).map(
             |(file_type, rights_base, rights_inheriting)| Self {
                 fd_object: FdObject {
@@ -99,11 +99,11 @@ impl FdEntry {
         )
     }
 
-    pub fn duplicate(file: &fs::File) -> Result<Self> {
+    pub(crate) fn duplicate(file: &fs::File) -> Result<Self> {
         Self::from(file.try_clone()?)
     }
 
-    pub fn duplicate_stdin() -> Result<Self> {
+    pub(crate) fn duplicate_stdin() -> Result<Self> {
         determine_type_and_access_rights(&io::stdin()).map(
             |(file_type, rights_base, rights_inheriting)| Self {
                 fd_object: FdObject {
@@ -118,7 +118,7 @@ impl FdEntry {
         )
     }
 
-    pub fn duplicate_stdout() -> Result<Self> {
+    pub(crate) fn duplicate_stdout() -> Result<Self> {
         determine_type_and_access_rights(&io::stdout()).map(
             |(file_type, rights_base, rights_inheriting)| Self {
                 fd_object: FdObject {
@@ -133,7 +133,7 @@ impl FdEntry {
         )
     }
 
-    pub fn duplicate_stderr() -> Result<Self> {
+    pub(crate) fn duplicate_stderr() -> Result<Self> {
         determine_type_and_access_rights(&io::stderr()).map(
             |(file_type, rights_base, rights_inheriting)| Self {
                 fd_object: FdObject {
--- a/src/fdentry.rs
+++ b/src/fdentry.rs
@@ -1,86 +1,40 @@
-use crate::sys::fdentry_impl;
+use crate::sys::fdentry_impl::{determine_type_and_access_rights, OsFile};
 use crate::{host, Error, Result};
-use cfg_if::cfg_if;
 use std::mem::ManuallyDrop;
 use std::path::PathBuf;
 use std::{fs, io};
 
-cfg_if! {
-    if #[cfg(any(
-        target_os = "macos",
-        target_os = "freebsd",
-        target_os = "netbsd",
-        target_os = "openbsd",
-        target_os = "ios",
-        target_os = "dragonfly"
-    ))] {
-        use std::sync::Mutex;
-
-        #[derive(Debug)]
-        pub struct DirStream {
-            pub file: ManuallyDrop<fs::File>,
-            pub dir_ptr: *mut libc::DIR,
-        }
-
-        impl Drop for DirStream {
-            fn drop(&mut self) {
-                unsafe { libc::closedir(self.dir_ptr) };
-            }
-        }
-
-        #[derive(Debug)]
-        pub enum Descriptor {
-            File {
-                file: fs::File,
-                dir_stream: Option<Mutex<DirStream>>,
-            },
-            Stdin,
-            Stdout,
-            Stderr,
-        }
-
-        impl From<fs::File> for Descriptor {
-            fn from(file: fs::File) -> Self {
-                Descriptor::File {
-                    file,
-                    dir_stream: None
-                }
-            }
-        }
-    } else {
-        #[derive(Debug)]
-        pub enum Descriptor {
-            File {
-                file: fs::File,
-            },
-            Stdin,
-            Stdout,
-            Stderr,
-        }
-
-        impl From<fs::File> for Descriptor {
-            fn from(file: fs::File) -> Self {
-                Descriptor::File { file }
-            }
-        }
-    }
+#[derive(Debug)]
+pub enum Descriptor {
+    OsFile(OsFile),
+    Stdin,
+    Stdout,
+    Stderr,
 }
 
 impl Descriptor {
-    pub fn as_file(&self) -> Result<&fs::File> {
+    pub fn as_file(&self) -> Result<&OsFile> {
         match self {
-            Descriptor::File { file, .. } => Ok(file),
+            Descriptor::OsFile(file) => Ok(file),
+            _ => Err(Error::EBADF),
+        }
+    }
+
+    pub fn as_file_mut(&mut self) -> Result<&mut OsFile> {
+        match self {
+            Descriptor::OsFile(file) => Ok(file),
             _ => Err(Error::EBADF),
         }
     }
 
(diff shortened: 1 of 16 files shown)


src/hostcalls_impl/fs.rs
```diff
@@ -25,6 +25,10 @@ pub(crate) fn fd_close(wasi_ctx: &mut WasiCtx, fd: wasm32::__wasi_fd_t) -> Resul
     }
 
     let mut fe = wasi_ctx.fds.remove(&fd).ok_or(host::__WASI_EBADF)?;
+
+    #[cfg(not(any(target_os = "linux", target_os = "windows")))]
```
marmistrz: I'd rather be explicit here
```
     #[cfg(any(target_os = "freebsd", target_os = "openbsd", target_os = "netbsd", target_os = "macos")))] 
```
in case we decide to support e.g. Android or Fuchsia in the future. 
The possible values of `target_os` can be found [here](https://docs.rs/platforms/0.2.0/platforms/target/enum.OS.html)

src/sys/unix/hostcalls_impl/fs.rs
```diff
@@ -268,6 +274,106 @@ pub(crate) fn fd_readdir(
     Ok(host_buf_len - left)
 }
 
+#[cfg(not(target_os = "linux"))]
```
marmistrz: I'd rather be explicit here. You could also use
```
#[cfg(...)]
mod fd_readdir_impl_bsd {
     ... 
}
#[cfg(...)]
pub use fd_readdir_impl_bsd::fd_readdir;
```
to reduce the number of #[cfg]s

src/sys/unix/hostcalls_impl/fs.rs
```diff
@@ -268,6 +274,106 @@ pub(crate) fn fd_readdir(
...
+lazy_static::lazy_static! {
+    static ref DIRP_CACHE: Arc<RwLock<HashMap<RawFd, AtomicPtr<libc::DIR>>>> = Arc::new(RwLock::new(HashMap::new()));
```
marmistrz: Please add a comment why the cache is needed.

marmistrz (changes requested):

alexcrichton:
Ah sorry I don't really know much about the system-level details here, so I don't think I'd be a great reviewer for this

sunfishcode:
Instead of a `HashMap<RawFd, *mut libc::DIR>` in a `lazy_static`, would it be possible to store the `*mut libc::DIR` in the `FdEntry`, `FdObject`, or `Descriptor`, whichever is most appropriate? It feels like the `*mut libc::DIR` should be just additional per-file-descriptor state.

kubkon:
@sunfishcode that’s an excellent idea, thanks!

src/sys/unix/hostcalls_impl/fs.rs
kubkon: I like this approach, thanks!

src/fdentry.rs
```diff
@@ -3,27 +3,70 @@ use crate::{host, Error, Result};
 
 use std::mem::ManuallyDrop;
 use std::path::PathBuf;
+#[cfg(any(
+    target_os = "macos",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "openbsd"
```
sunfishcode: Should these changes apply to "ios" and "dragonfly" too? I guess we can just make our best guess here.

src/sys/unix/hostcalls_impl/fs.rs
```diff
@@ -270,6 +273,96 @@ pub(crate) fn fd_readdir(
...
+    let (fd, dir_stream) = match fd {
+        Descriptor::File { file, dir_stream } => (file.as_raw_fd(), dir_stream),
+        _ => return Err(Error::EBADF),
+    };
+
+    let dir = if let Some(dirp) = dir_stream {
+        dirp.load(SeqCst)
+    } else {
+        let dirp = unsafe { fdopendir(fd) };
+        dir_stream.replace(AtomicPtr::new(dirp));
```
sunfishcode: A `replace` where you don't use the result is just a store.

src/fdentry.rs
```diff
@@ -3,27 +3,70 @@ use crate::{host, Error, Result};
...
-    File(fs::File),
+    File {
+        file: fs::File,
+        #[cfg(any(
+            target_os = "macos",
+            target_os = "freebsd",
+            target_os = "netbsd",
+            target_os = "openbsd"
+        ))]
+        dir_stream: Option<AtomicPtr<libc::DIR>>,
```
sunfishcode: Should this be a `Mutex` rather than `AtomicPtr`, to prevent multiple threads from doing `readdir` at the same time?

src/fdentry.rs
```diff
@@ -3,27 +3,70 @@ use crate::{host, Error, Result};
...
     Stdout,
     Stderr,
 }
 
+#[cfg(not(any(
+    target_os = "macos",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "openbsd"
+)))]
```
sunfishcode: `not(any(lots of stuff))` in a file which also does `any(lots of stuff)` suggests a `cfg_if` macro might be useful.

Alternatively, since code later does `#[cfg(target_os = "linux")]`, should we do that here too?

src/sys/unix/hostcalls_impl/fs.rs
```diff
@@ -270,6 +273,96 @@ pub(crate) fn fd_readdir(
...
+
+    let (fd, dir_stream) = match fd {
+        Descriptor::File { file, dir_stream } => (file.as_raw_fd(), dir_stream),
+        _ => return Err(Error::EBADF),
+    };
+
+    let dir = if let Some(dirp) = dir_stream {
+        dirp.load(SeqCst)
+    } else {
+        let dirp = unsafe { fdopendir(fd) };
```
marmistrz: You need to `dup` the fd for fdopendir, see the relevant source code comment in #71.

marmistrz (changes requested):

src/sys/unix/hostcalls_impl/fs.rs
kubkon: Good point! Thanks for catching that one!

src/fdentry.rs
kubkon: "ios" probably yes, "dragonfly" no idea. I've got least experience with the latter actually. But I guess we can add it in and indeed hope for the best ;-)

src/fdentry.rs
kubkon: I've now reshuffled the code a bit. Hopefully it's a bit cleaner. I'm not set on this organisation though, so if you feel it should be done in a different way, fire away please!

src/sys/unix/hostcalls_impl/fs.rs
kubkon: I've now made sure we clone the `fd` and then close it with `closedir` as prescribed in the man pages :-)

src/fdentry.rs
kubkon: I've now added both `ios` and `dragonfly` to the list.

src/sys/unix/bsd/hostcalls_impl.rs
```diff
@@ -0,0 +1,108 @@
...
+    _offset: host::__wasi_filesize_t,
+    _len: host::__wasi_filesize_t,
+) -> Result<()> {
+    match advice {
+        host::__WASI_ADVICE_DONTNEED
+        | host::__WASI_ADVICE_SEQUENTIAL
+        | host::__WASI_ADVICE_WILLNEED
+        | host::__WASI_ADVICE_NOREUSE
+        | host::__WASI_ADVICE_RANDOM
+        | host::__WASI_ADVICE_NORMAL => {}
```
sunfishcode, line 151: This is pre-existing, but least some BSD's [do support `posix_fadvise`](https://www.freebsd.org/cgi/man.cgi?query=posix_fadvise&sektion=2&manpath=FreeBSD+8.3-RELEASE), so we should have a TODO comment here.

src/sys/unix/linux/hostcalls_impl.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    let mut entry_buf = unsafe { std::mem::uninitialized::<dirent>() };
+    let mut left = host_buf_len;
+    let mut host_buf_offset: usize = 0;
+    while left > 0 {
+        let mut host_entry: *mut dirent = std::ptr::null_mut();
+        let res = unsafe { readdir_r(dir, &mut entry_buf, &mut host_entry) };
+        if res == -1 {
+            return Err(host_impl::errno_from_nix(nix::errno::Errno::last()));
+        }
```
sunfishcode, line 45: This is pre-existing, but I just noticed it now. `readdir_r` returns a positive int on failure, and doesn't set errno.

It would also be good to look into using plain `readdir` instead of `readdir_r`, since [the Linux readdir_r documentation](http://man7.org/linux/man-pages/man3/readdir_r.3.html) says `readdir_r` is deprecated, though we can do that later if you just want to leave a TODO comment or so.

src/sys/unix/bsd/osfile.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+impl Drop for DirStream {
+    fn drop(&mut self) {
+        unsafe { libc::closedir(self.dir_ptr) };
+    }
+}
+
+#[derive(Debug)]
+pub struct OsFile {
```
sunfishcode: Can this (and the other `OsFile` implementations, and `DirStream`, be `pub(crate)` to make it clear that this isn't part of the public API?

src/sys/unix/bsd/hostcalls_impl.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        let mut entry: host::__wasi_dirent_t =
+            host_impl::dirent_from_host(&unsafe { *host_entry })?;
+        // Set d_next manually:
+        // * on macOS d_seekoff is not set for some reason
+        // * on FreeBSD d_seekoff doesn't exist; there is d_off but it is
+        //   not equivalent to the value read from telldir call
+        entry.d_next = unsafe { telldir(dir_stream.dir_ptr) } as host::__wasi_dircookie_t;
+
+        log::debug!("entry = {:?}", entry);
```
sunfishcode: This looks like a useful `debug!` macro; would you mind adding a similar one in the Linux code?

src/sys/unix/linux/hostcalls_impl.rs
```diff
@@ -0,0 +1,94 @@
...
+            memcpy(
+                host_buf_ptr.offset(host_buf_offset as isize) as *mut _,
+                name_ptr as *const _,
+                name_len,
+            )
+        };
+        host_buf_offset += name_len;
+        left -= required_space;
+    }
+    Ok(host_buf_len - left)
```
sunfishcode, line 76: This is a minor nit, but the BSD version has a blank line before the `Ok` here, and it's nice to keep the formatting consistent.

sunfishcode (commented):
Looks good! Just a few more minor comments:

src/sys/unix/bsd/hostcalls_impl.rs
kubkon: Done!

src/sys/unix/linux/hostcalls_impl.rs
kubkon: Done!

src/sys/unix/bsd/hostcalls_impl.rs
kubkon: Done!

src/sys/unix/linux/hostcalls_impl.rs
kubkon: Done!

src/fdentry.rs
```diff
@@ -1,47 +1,56 @@
...
 
+    #[allow(unused)]
     pub fn is_stdout(&self) -> bool {
         match self {
             Descriptor::Stdout => true,
             _ => false,
         }
     }
 
+    #[allow(unused)]
```
marmistrz, line 53: Add a comment explaining this supression

src/hostcalls_impl/fs.rs
```diff
@@ -203,7 +203,7 @@ pub(crate) fn fd_renumber(
         .fd_object
         .descriptor
         .as_file()
-        .and_then(FdEntry::duplicate)?;
+        .and_then(|file| FdEntry::duplicate(file))?;
```
marmistrz, line 206: Why is the closure neeeed here?

src/sys/unix/bsd/hostcalls_impl.rs
```diff
@@ -0,0 +1,153 @@
...
+    if os_file.dir_stream.is_none() {
+        let file = os_file.file.try_clone()?;
+        let dir_ptr = unsafe { fdopendir(file.as_raw_fd()) };
+        os_file.dir_stream = Some(Mutex::new(DirStream {
+            file: ManuallyDrop::new(file),
+            dir_ptr,
+        }));
+    }
+    let dir_stream = os_file.dir_stream.as_mut().unwrap().lock().unwrap();
+
```
marmistrz, line 32: expect? or add a comment why unwrap is ok here.

src/sys/unix/bsd/hostcalls_impl.rs
marmistrz: mention  fd_readdir in this `debug!`

src/sys/unix/bsd/hostcalls_impl.rs
```diff
@@ -0,0 +1,153 @@
...
+        log::debug!("entry = {:?}", entry);
+
+        let name_len = entry.d_namlen as usize;
+        let required_space = std::mem::size_of_val(&entry) + name_len;
+        if required_space > left {
+            break;
+        }
+        unsafe {
+            let ptr = host_buf_ptr.offset(host_buf_offset as isize) as *mut c_void
```
marmistrz: try_into?

src/sys/unix/bsd/osfile.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub file: fs::File,
+    pub dir_stream: Option<Mutex<DirStream>>,
```
marmistrz: Nitpicking: or RwLock?

marmistrz (changes requested):
Also, add a comment explaining the purpose of `OsFile` and the scope where it should be used

src/hostcalls_impl/fs.rs
kubkon: Good question. `FdEntry::duplicate` requires `&fs::File` whereas currently, within a `FdEntry::FdObject::Descriptor` we store an `OsFile` which implements a `Deref<Target = fs::File` but it's not enough to convince the compiler that it can deref `file` in the call above into `&fs::File`, hence why the explicit closure here. :-)

src/sys/unix/bsd/hostcalls_impl.rs
kubkon: Comment added!

src/sys/unix/bsd/hostcalls_impl.rs
kubkon: Done!

src/sys/unix/bsd/hostcalls_impl.rs
marmistrz: How about `unwrap_or_else` instead of the if?

kubkon: Yeah, I thought about it. I also considered `get_or_insert_with`, but alas, neither of these allow for the closure to fail with an error, and we need this since we need to duplicate the `File` handle as you rightly pointed out in one of your previous comments. Anyhow, if you do know a way, fire away and I'll be more than happy to change it! :-)

src/fdentry.rs
kubkon: The suppression is there since we actually don't rely on any of this functionality in the lib just yet. This is not to say we definitely will in the future, but it makes the interface of `Descriptor` a little bit more complete. I can either leave the suppression in or simply chuck away the unused `fn`s. I guess personally I'd leave the suppression in for now and decide later if it's needed or not :-)

src/sys/unix/bsd/osfile.rs
kubkon: Yep, good catch, thanks! I've changed it to public-private now :-)

src/sys/unix/bsd/hostcalls_impl.rs
marmistrz: `.map(Ok).unwrap_or_else(...)?`

src/sys/unix/bsd/osfile.rs
kubkon: Ah, now this one's tricky. I actually thought about using `RwLock` but then convinced myself otherwise as AFAICS we'll always require a write lock on the stream since every call to *dir syscall modifies the underlying `dir_ptr`.

src/sys/unix/bsd/hostcalls_impl.rs
kubkon: Hmm, interesting. Could it be used with `get_or_insert_with`? I just remembered that we cannot use `unwrap_or_else` as we need to hold onto the duplicated handle and the dir stream ptr after it's first created.

marmistrz: If you can do an unwrap, you can also do unwrap_or_else. You can probably mutate the option in unwrap_or_else. Alternatively, pattern match with mutation on None?

kubkon: Hmm, in that case I think I'd prefer a pattern match from the readability perspective. All things considered though, no solution here seems particularly elegant unfortunately.

kubkon: OK, I've changed it to pattern matching with mutation on `None`. Is that what you've had in mind?

marmistrz: Apples & oranges again, for me unwrap or else conveys the meaning more clearly.

marmistrz: it's ok, but use an expect for the mutex lock

kubkon: Good idea, so how about it we push it even further and throw an actual error instead of panicking? I was thinking of `ENOLCK`.

kubkon: I've added `Error::LockError` which resolves to `ENOLCK` :-) Have a look!

sunfishcode: The only thing `ENLCK` is used for in POSIX is `fcntl` file locking, so it's not necessarily the best fit here.

`Mutex`es return `PoisonError` after a thread holding the lock panics. The documentation says [it's common to use `unwrap()`](https://doc.rust-lang.org/std/sync/struct.Mutex.html#poisoning) when unlocking, because that effectively means that the panic is propagated to the other thread. I think that's what we want here. When the error occurs, there's already been a panic, which means there's a bug in wasi-common or an upstream dependency, so I think it's appropriate to keep panicking on any thread still trying to access the same `WasiCtx`.

BTW, thinking about threading issues here has me thinking that we should put more of `WasiCtx` behind a `Mutex` or `RwLock`. Right now, lots of WASI functions, like `fd_write`, take an `&mut WasiCtx`, which in Rust means that they should have exclusive access to it. However, nothing in WASI or Wasm prevents multiple threads from calling `fd_write` at the same time. Normally, Rust's borrow checker would protect us, but in this case, the calls may be coming from JIT code :-). But that's obviously outside the scope of this PR.

kubkon: Ah, that's good to know! Thanks for elaborating on it :-) I've now reverted the latest commit.

As far as threading issues are concerned, you've raised some interesting points especially since I believe I remember seeing some `Mutex`es and/or `RwLock`s in the original `wasmtime-wasi-c` implementation, so I guess revisiting that in subsequent PRs might indeed be a good idea :-)

kubkon:
I reckon the PR is in a pretty good state, and I'm inclined to merge it as is. Should any more issues/suggestions come to mind, let's address them in subsequent PRs :-)

PR #84, opened 2019-09-08
Refactor error handling

marmistrz:
Most of the interesting stuff happens in src/error.rs. No more need for `errno_from_ioerror`, `errno_from_win`, manual handling of `try_into` errors. 

There are still some remaining references to `errno_from_nix`, but intend to remove them in a subsequent PR.

Let me know if I should implement a better `Display` for `Error`.

TODO:
* [x] implement `Fail`/`std::error::Error` for `Error`


src/error.rs
```diff
@@ -0,0 +1,265 @@
...
+    ERANGE = 68,
+    EROFS = 69,
+    ESPIPE = 70,
+    ESRCH = 71,
+    ESTALE = 72,
+    ETIMEDOUT = 73,
+    ETXTBSY = 74,
+    EXDEV = 75,
+    ENOTCAPABLE = 76,
+}
```
sunfishcode, line 87: Would it work to reference the constants in host.rs instead of duplicating them here? Or, does this change mean we can remove the constants from host.rs?

sunfishcode (commented):
Nice!

src/error.rs
marmistrz: > Would it work to reference the constants in host.rs instead of duplicating them here? 

Good idea! Done.

> Or, does this change mean we can remove the constants from host.rs?

Not yet, we're still doing comparisons with constants from host.rs in a couple of places. But hopefully yes in the future.

kubkon (approved):
LGTM!

PR #87, opened 2019-09-08
Update to syn 1 and quote 1.

sunfishcode:
I was doing some experiments with the wasi-common-cbindgen macro and ran into some limitations of the quote macro. Looking into it, I found they were fixed in a newer version. This PR just updates to the newer version.


kubkon (approved):
LGTM!

PR #89, opened 2019-09-09
Make functions that operate on raw I/O handles unsafe.

sunfishcode:
Functions which trust that their arguments are valid raw file descriptors
or raw handles should be marked unsafe, because these arguments are
passed unchecked to I/O routines.


kubkon:
LGTM!

PR #90, opened 2019-09-09
Mark public API functions as unsafe.

sunfishcode:
This marks the public hostcalls functions as unsafe.

This is generalizing from Rust's `from_raw_fd` function, which is
unsafe. The observation is that nothing prevents code using this
function from passing a bogus or stale dangling file descriptor and
corrupting an arbitrary open stream.

Technically, some of these functions don't use file descriptors, such as
random, clocks, and a few others. However I expect that in the future,
random and clocks will switch to using file descriptors anyway, and it
keeps the macro definitions simpler if we only have to handle one form.

This is a complex topic, so I'm not in a hurry here. I want to figure out
what the right thing to do is.

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -163,11 +163,11 @@ impl WasiCtx {
             .and_then(|ctx| ctx.build())
     }
 
-    pub(crate) fn contains_fd_entry(&self, fd: host::__wasi_fd_t) -> bool {
+    pub(crate) unsafe fn contains_fd_entry(&self, fd: host::__wasi_fd_t) -> bool {
         self.fds.contains_key(&fd)
     }
 
-    pub(crate) fn get_fd_entry(
+    pub(crate) unsafe fn get_fd_entry(
         &self,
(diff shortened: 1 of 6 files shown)


kubkon (approved):
Looks good! Question though: should we make host-specific hostcalls impl functions `unsafe` as well? I'm thinking here about `hostcalls_impl` modules in `sys` module.

sunfishcode:
> should we make host-specific hostcalls impl functions unsafe as well? I'm thinking here about hostcalls_impl modules in sys module.

Right now, by mental model is that anything that takes a raw file descriptor as an argument should be `unsafe`, so no, most of those functions are fine.

kubkon:
> > should we make host-specific hostcalls impl functions unsafe as well? I'm thinking here about hostcalls_impl modules in sys module.
> 
> Right now, by mental model is that anything that takes a raw file descriptor as an argument should be `unsafe`, so no, most of those functions are fine.

Sounds good!

Issue #83, opened 2019-09-07
[Brainstorming] High-level API based on wasi-common?


sunfishcode:
Would it make sense to create a high-level `std::fs`-like API, but capability-based, on top of wasi-common? I'm not aware of a present need for such a thing, so this is just brainstorming, but I'm interested if others have thoughts on this.

It is interesting that [Rust’s `from_raw_fd` is `unsafe`](https://doc.rust-lang.org/std/os/unix/io/trait.FromRawFd.html#tymethod.from_raw_fd). It turns out that this is an [arbitrary choice](https://github.com/CraneStation/rust-wasi/issues/8#issuecomment-525711065). Rust could have decided that any code can take any integer and do any I/O operation on it (on unix-style platforms), and it wouldn’t lead to any memory unsafety or any undefined behavior. But instead, Rust chose for safe to mean more than just a lack of undefined behavior; it also applies to I/O.

This means that once you have an open I/O stream, the only way to access that stream (without `unsafe`) is through the Rust object representing it. This has close parallels to a capability-based security model. However, Rust still lets you open files by absolute path.

So what if we made a high-level API that was similar to `std::fs`, but capability-based, on top of wasi-common? Instead of `File::open` and `read_dir` being standalone functions (ambient authority) that take `Path`s, there'd be a `Dir` struct representing a directory handle, with methods like `Dir::open_file` and `Dir::open_dir`  that only open files and directories relative to and within that directory, and a `Dir::read` which would return a `ReadDir` for iterating over its contents.

jedisct1:
That would be very useful and make common operations easier.

But this is something that, eventually, will have to be written for every single programming language.

We had to do something similar for AssemblyScript already. Because exposing only file descriptors is neither idiomatic nor convenient.

The problem with everybody reimplementing high-level wrappers is that they may end up being completely different from one implementation to another.

Having implementations explicitly described as reference implementations may help having something a little more uniform, at least for how functions and structures are named.

As a side note, at least for filesystem operations, I really wish we adopted mechanisms similar to OpenBSD's `unveil(2)` instead of pre-opened descriptors. If only because application themselves can easily restrict the filesystem's view as needed. For other capabilities, an equivalent to `pledge(2)` would also be a great step forward. Because once again, applications know when they can drop capabilities, and which ones.

sunfishcode:
[...]

PR #91, opened 2019-09-10
Begin sketching out a new high-level `fs` API.

sunfishcode:
This is a very preliminary sketch of #83. It doesn't even compile yet, but it shows a possible high-level structure of such an API.


src/fs/ctx.rs
```diff
@@ -0,0 +1,7 @@
+use crate::WasiCtx;
+
+lazy_static! {
+    // TODO: Should we allow the context to be passed alternate arguments?
```
kubkon: Hmm, I guess we should. How otherwise are we going to pass in the preopens? Or are we going to add another mid-level call for that perhaps?

sunfishcode: Thinking about this a little more, I don't think a `lazy_static` singleton is the way to go, because people using this API may want to have multiple distinct `WasiCtx`s that are isolated from each other. I've now updated the patch to have `Dir` and `File` hold a reference to a `WasiCtx`. We may further want to make it an `Rc`, but we'll see how things go.

sunfishcode:
This is now complete enough to compile, and a lot of the API surface area is now in place, though most of the functionality isn't implemented yet.

src/fs/ctx.rs
kubkon: Good point!

src/fs/dir.rs
```diff
@@ -0,0 +1,216 @@
...
+            path.as_os_str().len(),
+            host::__WASI_O_CREAT | host::__WASI_O_TRUNC,
+            !0,
+            !0,
+            0,
+            &mut fd,
+        ))?;
+        */
+
+        let ctx = self.ctx;
```
kubkon, line 146: Is this necessary here? Could we not simply pass `self.ctx` directly?

src/fs/error.rs
```diff
@@ -0,0 +1,265 @@
...
+        | host::__WASI_ENOTCAPABLE => {
+            return Err(io::Error::new(io::ErrorKind::Other, error_str(errno)))
+        }
+        _ => panic!("unrecognized WASI errno value"),
+    } as i32;
+
+    Err(io::Error::from_raw_os_error(raw_os_error))
+}
+
+#[cfg(windows)]
```
kubkon, line 184: Do you think that in the future it'd be useful to have the same setup as for the hostcalls? That is, separate host specifics in modules and then just do a `cfg_if` instead of multiple `#[cfg(..)]`s? Mind you, this is an idea for the future and not now; I'm happy with the state of `error.rs` as is :-)

src/fs/metadata.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// Metadata information about a file.
+///
+/// This corresponds to [`std::fs::Metadata`].
+///
+/// TODO: Not yet implemented.
+///
+/// [`std::fs::Metadata`]: https://doc.rust-lang.org/std/fs/struct.Metadata.html
+#[derive(Clone)]
+pub struct Metadata {}
```
kubkon, line 12:
```suggestion
pub struct Metadata;
```

kubkon (approved):
Looks great! Here's a thought: could we actually re-use this interface and the implementation in Rust's `libstd`? I might be way off base here but browsing through the changes here looks as if it could be re-used in `libstd`. @alexcrichton thoughts?

alexcrichton:
Seems possible perhaps! I suspect this'd want to bake a bit more before going into libstd, but the internal details of libstd are up for grabs to change at any time!

sunfishcode:
Interesting idea. In addition to obviously baking :-), a few additional pieces we'd need:

 - A way for this `fs` API to use WASI directly, instead of going through the current wasi-common userspace implementation. Some of that may be as simple as just adding alternate implementations of the API under `#[cfg(target_os = wasi)]`. What should we do about `WasiCtx` though? In wasi-common, one can create multiple independent `WasiCtx`s and they're explicitly passed into the API. In WASI itself, there's currently just one implicit `WasiCtx` (or equivalent) below the platform boundary. One option is to change WASI itself to have a context argument and support independent contexts, I suppose.

 - Something similar to what `libpreopen` does within WASI libc, so that when Rust code calls `File::open("/absolute/path")`, we can remap it to directory fd and relative path. This should be straight-forward.

 - "Rust-idiomatic" wrappers on top of wasi-common's low-level API, similar to [the wasi crate](https://github.com/CraneStation/rust-wasi/), using `Result`s and slices rather than bare errnos and pointers. This is what the Rust libstd support for WASI [currently uses](https://github.com/rust-lang/rust/pull/63676), and it would be nice if this new `fs` layer could use something similar.

@cc newpavlov since this is one possible path to a libc-free libstd for WASI targets.

sunfishcode:
cc @newpavlov even, since this is one possible path to a libc-free libstd for WASI targets.

newpavlov:
>this is one possible path to a libc-free libstd for WASI targets.

I am not sure if I understand how it will work. All WASI code still works via Core API at the lowest level, right? And this high-level API is still built on top of Core API?

Right now, we depend on `libc` in `std` only for interoperability with C/C++. Removing the remaining `libc` usages should be fairly trivial if we'll add a separate target (e.g. `wasm32-wasi-rust`).

Not strictly relevant, but can you explain how `__wasilibc_find_relpath` works in terms of Core API? I am not completely sure, but it looks like we can remove its usage from `std`.

sunfishcode:
This builds, and although most of the functionality is not yet implemented, I'd like to land it and continue to iterate incrementally.

kubkon:
@sunfishcode sounds great to me! Feel free to merge it or I can do that if you want. Also, apologies for little to no activity lately but got swamped with preparation work for the upcoming Devcon5 conference :-)

PR #92, opened 2019-09-11
Mark `WasiCtx`'s internals as `pub(crate)`.

sunfishcode:
Users of wasi-common don't need to be manipulating `WasiCtx`'s internals directly.


kubkon (approved):
Agreed! :-)

PR #94, opened 2019-09-11
Add a method for inhering the command-line arguments from the host.

sunfishcode:
For convenience in the case where one wants to forward the host command-line arguments to the `WasiCtx`.


kubkon (approved):
LGTM!

PR #96, opened 2019-09-11
Quick fix for path_filestat_set_times. Closes #95.

### LINKED ISSUE #95: SET_ATIM in path_filestat_set_times
The if-else chain here:

https://github.com/CraneStation/wasi-common/blob/master/src/sys/unix/hostcalls_impl/fs.rs#L426

checks for `set_atim_now` twice. It looks like the first condition should check `set_atim` instead, similar to the `mtim` version below it.

marmistrz:


sunfishcode (approved):
This looks right to me! Would you mind writing a test to path_filestat.rs to cover at least `SET_ATIM`? Writing a robust test for the `_NOW` cases seems like it might be tricky, so I'm ok making a note of that somewhere and deferring that for now.

marmistrz:
@sunfishcode please make an issue and assign it to me. My availability in the second half of September is rather limited, so I'll extend the test when I have while. I have some idea how to rudimentarily test `*_NOW`, I'll implement both test cases together.

PR #101, opened 2019-09-16
Refactor mgmt of misc testsuite

kubkon:
Changes:
* use [tempfile] crate for auto mgmt of temp dirs
* use concrete types in place of generics in `utils` module

[tempfile]: https://github.com/Stebalien/tempfile

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -113,7 +113,7 @@ fn write_testsuite_tests(out: &mut File, dir_entry: DirEntry, testsuite: &str) -
         "    fn {}() -> Result<(), String> {{",
         avoid_keywords(&stemstr.replace("-", "_"))
     )?;
-    write!(out, "        setup_log();")?;
+    writeln!(out, "        setup_log();")?;
     write!(out, "        let path = std::path::Path::new(\"")?;
     // Write out the string with escape_debug to prevent special characters such
     // as backslash from being reinterpreted.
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -129,7 +129,11 @@ fn write_testsuite_tests(out: &mut File, dir_entry: DirEntry, testsuite: &str) -
     let workspace = if no_preopens(testsuite, stemstr) {
         "None"
     } else {
-        "Some(&utils::prepare_workspace(&bin_name)?)"
+        writeln!(
+            out,
+            "        let workspace = utils::prepare_workspace(&bin_name)?;"
+        )?;
+        "Some(workspace.path())"
```
sunfishcode, line 136: The switch to `TempDir` is good; while we're here, would it be possible to keep the `TempDir` object live, rather than dropping it here? This is a theoretical concern, but the second time we create the directory, we treat the case where the path already somehow exists as an error. I don't think it's insecure, but it could lead to spurious errors in very rare cases.

sunfishcode (commented):
Good idea!

build.rs
kubkon: Hmm, could you elaborate a bit more? Do you mean having one global static `TempDir` ref? If so, I've tried that already, I just couldn't find a way to have the "destructor" for the static called automatically after all tests are run (which is purposely not run for any statics in Rust and for a good reason AFAIK). This basically implies that the auto-cleanup functionality of `tempfile` crate is not utilised, but this is not that critical if you ask me.

sunfishcode: You're right, this isn't critical. Let's merge this, and I may poke around to see if I can implement what I'm thinking of.

sunfishcode: Actually, looking at this further, I think I was just mistaken about this. So we're good here!

PR #102, opened 2019-09-17
Fix path_remove_directory on *nixes

kubkon:
This PR fixes `path_remove_directory` on *nix hosts which includes Linux and macOS. Windows, however, still needs more work in order to make `remove_directory_trailing_slashes` pass.

PR #103, opened 2019-09-17
Fix path_rename on *nix hosts

kubkon:
The fix contains an errno remapping in macOS case where in case when we try to rename a file into a path with a trailing slash an `ENOENT` is returned. In this case, if the destination does not exist, an `ENOTDIR` should be thrown as is thrown correctly on Linux hosts. Thus, as a fix, if an `ENOENT` is thrown, an additional check is performed to see whether the destination path indeed contains a trailing slash, and if so, the errno is adjusted to `ENOTDIR` to match the POSIX/WASI spec.

Alas, Windows requires more investigation in order to provide a fix, so it's currently skipped.

This PR builds on #102 hence it is expected for the former to be merged first before we merge this PR.

### DIFF
--- a/src/sys/unix/bsd/hostcalls_impl.rs
+++ b/src/sys/unix/bsd/hostcalls_impl.rs
@@ -1,10 +1,60 @@
 use super::osfile::OsFile;
+use crate::hostcalls_impl::PathGet;
 use crate::sys::host_impl;
 use crate::{host, Error, Result};
 use nix::libc::{self, c_long, c_void};
 use std::convert::TryInto;
+use std::ffi::CString;
 use std::fs::File;
 use std::os::unix::prelude::AsRawFd;
+
+pub(crate) fn path_rename(resolved_old: PathGet, resolved_new: PathGet) -> Result<()> {
+    use nix::{errno::Errno, fcntl::AtFlags, libc::renameat, sys::stat::fstatat};
+    let old_path_cstr = CString::new(resolved_old.path().as_bytes()).map_err(|_| Error::EILSEQ)?;
+    let new_path_cstr = CString::new(resolved_new.path().as_bytes()).map_err(|_| Error::EILSEQ)?;
+
+    let res = unsafe {
+        renameat(
+            resolved_old.dirfd().as_raw_fd(),
+            old_path_cstr.as_ptr(),
+            resolved_new.dirfd().as_raw_fd(),
+            new_path_cstr.as_ptr(),
+        )
+    };
+    if res != 0 {
+        // Currently, this is verified to be correct on macOS, where
+        // ENOENT can be returned in case when we try to rename a file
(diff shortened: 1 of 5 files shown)


src/sys/unix/linux/hostcalls_impl.rs
```diff
@@ -1,11 +1,33 @@
...
 use nix::libc::{self, c_long, c_void};
 use std::convert::TryInto;
+use std::ffi::CString;
 use std::fs::File;
 use std::os::unix::prelude::AsRawFd;
 
+pub(crate) fn path_rename(resolved_old: PathGet, resolved_new: PathGet) -> Result<()> {
+    use nix::libc::renameat;
+    let old_path_cstr = CString::new(resolved_old.path().as_bytes()).map_err(|_| Error::EILSEQ)?;
+    let new_path_cstr = CString::new(resolved_new.path().as_bytes()).map_err(|_| Error::EILSEQ)?;
```
sunfishcode, line 14: As an aside, this sequence of converting to a `CString` and mapping errors to `EILSEQ` happens a lot, so it would make for a good utility function.

src/sys/unix/bsd/hostcalls_impl.rs
```diff
@@ -1,11 +1,53 @@
...
+        // correct behaviour of throwing an ENOTDIR since the destination is
+        // indeed not a directory.
+        //
+        // TODO
+        // Verify on other BSD-based OSes.
+        Err(match Errno::last() {
+            Errno::ENOENT => {
+                // check if destination contains a trailing slash
+                if resolved_new.path().contains('/') {
+                    Error::ENOTDIR
```
sunfishcode: If the old path doesn't exist, we should still return `ENOENT` here. I know that requires making additional system calls here, but this should be a relatively cold path.

src/sys/unix/linux/hostcalls_impl.rs
kubkon: Excellent point! I've now added that to my todo list :-)

kubkon: And I've filed an issue to track this #104 :-)

src/sys/unix/bsd/hostcalls_impl.rs
kubkon: Ah, good point, thanks for catching it!

kubkon: Done!

kubkon:
@sunfishcode do you think we can merge it, or is there anything else you'd like fixed?

kubkon:
In order not to impede progress on anything else, I'll go ahead and merge this PR then if there are no more comments/suggestions. If you find anything amiss with the merged changes, feel free to open an issue :-)

sunfishcode:
Yes; this looks good!

PR #107, opened 2019-10-02
Update dependencies and use Once::new() instead of ONCE_INIT

kubkon:
This change is [Reviewable](https://reviewable.io/reviews/cranestation/wasi-common/107)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,11 +27,11 @@ winx = { path = "winx" }
 winapi = "0.3"
 
 [dev-dependencies]
-wasmtime-runtime = { git = "https://github.com/cranestation/wasmtime", rev = "042c877" }
-wasmtime-environ = { git = "https://github.com/cranestation/wasmtime", rev = "042c877" }
-wasmtime-jit = { git = "https://github.com/cranestation/wasmtime", rev = "042c877" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -15,9 +15,9 @@ wasi-common-cbindgen = { path = "wasi-common-cbindgen" }
 failure = "0.1"
 libc = "0.2"
 rand = "0.6"
-cfg-if = "0.1.9"
+cfg-if = "0.1"
 log = "0.4"
-filetime = "0.2.7"
+filetime = "0.2"
```
sunfishcode: I see projects that include minor versions, and projects that exclude them, but I'm unclear on what the motivations for each way are.

kubkon: Yeah, I'm a bit hazy on this topic as well. It looks to me however that binaries tend to include minor versions mostly, whereas libs tend to omit them. But if you reckon that's an issue, I'll happily revert back to minor versions.

kubkon: OK, I've reverted versioning back to its original state mainly due to the fact that dependabot has already updated a couple of dependencies ;-)

sunfishcode: Oops, I forgot to post here. I asked for advice, and people suggested we enable dependabot and leave the minor versions in. I don't have a strong opinion on this myself, so if there are reasons to do something different, let's talk :-).

PR #113, opened 2019-10-03
Misc testsuite feature gated

kubkon:
This PR puts building and generating of misc_testsuite behind a feature gate "misc_testsuite". This is mainly to allow projects which pull `wasi-common` as a dependency not to have to have `wasm32-wasi` target installed in order to build it as it currently is.

An unfortunate example which showcases the problems with our current setup can be found in CraneStation/wasmtime#398, where updating git rev of `wasi-common` used in `wasmtime` leads to build errors in the CI.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,9 @@ license = "Apache-2.0 WITH LLVM-exception"
 description = "WASI implementation in Rust"
 
 [features]
-misc_testsuite = []
+# this feature requires wasm32-wasi target installed, and it enables wasm32
+# integration tests when run with `cargo test --features wasm_tests`
+wasm_tests = []
 
 [dependencies]
 wasi-common-cbindgen = { path = "wasi-common-cbindgen" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ edition = "2018"
 license = "Apache-2.0 WITH LLVM-exception"
 description = "WASI implementation in Rust"
 
+[features]
+misc_testsuite = []
+
 [dependencies]
 wasi-common-cbindgen = { path = "wasi-common-cbindgen" }
 failure = "0.1"
--- a/README.md
+++ b/README.md
@@ -39,6 +39,28 @@ In our Windows implementation, we currently support the minimal subset of [WASI 
 which allows for running the very basic "Hello world!" style WASM apps. More coming shortly,
 so stay tuned!
 
+## Development hints
+When testing the crate, you may want to enable and run full wasm32 integration testsuite. This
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -10,6 +10,9 @@ edition = "2018"
 license = "Apache-2.0 WITH LLVM-exception"
 description = "WASI implementation in Rust"
 
+[features]
+misc_testsuite = []
```
sunfishcode: Could you add a comment here mentioning the need to install the wasm32-wasi target?

sunfishcode: Also as a minor bikeshed, we may have additional wasm testsuites in the future, so would it make sense to name this feature "wasm_tests"?

sunfishcode (approved):
Makes sense. Thanks for fixing this!

marmistrz (changes requested):
Please mention it in the README that one needs to pass `--features` to run integration tests. This may confuse new contributors.

Cargo.toml
kubkon: > Could you add a comment here mentioning the need to install the wasm32-wasi target?

Done!

> Also as a minor bikeshed, we may have additional wasm testsuites in the future, so would it make sense to name this feature "wasm_tests"?

Renaming to `wasm_tests` is a good idea! I was actually on the fence whether we should keep it as `misc_tests`/`misc_testsuite` and when we add more tests, have each "set" behind a separate feature gate, but I guess that for now, having one feature is more than enough. :-)

kubkon:
> Please mention it in the README that one needs to pass `--features` to run integration tests. This may confuse new contributors.

That's a good idea, thanks!

README.md
```diff
@@ -39,6 +39,28 @@ In our Windows implementation, we currently support the minimal subset of [WASI
...
+Next initiate submodules containing the integration testsuite
+
+```
+git submodule update --init
+```
+
+Now, you should be able to run the integration testsuite by enabling the `wasm_tests` feature
+
+```
+cargo test --features wasm_tests
```
marmistrz, line 61: Probably `rustup run ...`, if we're also describing the rustup steps.

kubkon: Uhm, I'm not sure I follow. I only included `rustup` as an example of how an additional target (here `wasm32-wasi`) can be installed. The rest is pure `cargo` only. Could you explain more what you mean by `rustup run ...` and possibly provide an illustrative example?

kubkon: Anyhow, if you don't mind, I'll go ahead and merge the PR as is so that CraneStation/wasmtime#398 lands :-)

PR #115, opened 2019-10-03
 Add a typesafe enum for file type 

marmistrz:
Depends on #114.

### DIFF
--- a/src/hostcalls_impl/fs.rs
+++ b/src/hostcalls_impl/fs.rs
@@ -1062,3 +1062,23 @@ pub(crate) unsafe fn fd_prestat_dir_name(
             enc_slice_of(memory, path.as_bytes(), path_ptr)
         })
 }
+
+#[allow(dead_code)] // trouble with sockets
+#[derive(Clone, Copy, Debug)]
+#[repr(u8)]
+pub(crate) enum FileType {
+    Unknown = host::__WASI_FILETYPE_UNKNOWN,
+    BlockDevice = host::__WASI_FILETYPE_BLOCK_DEVICE,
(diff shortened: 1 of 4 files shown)


src/hostcalls_impl/fs.rs
```diff
@@ -1062,3 +1062,22 @@ pub(crate) unsafe fn fd_prestat_dir_name(
...
+pub(crate) enum FileType {
+    Unknown = 0,
+    BlockDevice = 1,
+    CharacterDevice = 2,
+    Directory = 3,
+    RegularFile = 4,
+    SocketDgram = 5,
+    SocketStream = 6,
+    Symlink = 7,
+}
```
sunfishcode, line 1078: I like the idea here, but I think it would be good to keep "magic numbers" like this out of the main source files. Could these be defined in terms of constants like `__WASI_FILETYPE_CHARACTER_DEVICE` instead of hard-coding the numbers here? Or possibly, could we move this enum out of the `hostcalls_impl` directory?

marmistrz: Yes, of course, you're right about the magic numbers. I forgot to change it. Where would you best see this type? For me, a new file, `src/types.rs` sounds like the best option here.

marmistrz: For now, I have just fixed the magic numbers.

kubkon: I'll merge it as-is for now, and in the meantime, we can figure out how to re-organise the crate, or better yet, how about we do some re-org after we close off all missing hostcalls on Windows and remaining, known bugs?

kubkon (approved):

PR #116, opened 2019-10-03
Implement fd_readdir

marmistrz:
Obsoletes & closes #71. Depends on #114 & #115.

This is a redo of #71 - it was easier to redo the change basing on current master than to fix the merge conflicts. Most of the remarks from the previous PR will still be valid, but I'll state them in a subsequent comment for completeness.

@kubkon in a subsequent PR, would you like to either (a) adapt the forked `Dir` to work on *BSD so that we can reuse the code between OSes (b) expose a function returning `Vec<Entry>` so that we can reuse just the part consisting of copying data around


src/sys/unix/dir.rs
```diff
@@ -0,0 +1,291 @@
...
+///      descriptor. Each iteration rewinds when finished.
+///    * returns entries for `.` (current directory) and `..` (parent directory).
+///    * returns entries' names as a `CStr` (no allocation or conversion beyond whatever libc
+///      does).
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct Dir(ptr::NonNull<libc::DIR>);
+
+impl Dir {
+    /// Opens the given path as with `fcntl::open`.
+    pub fn open<P: ?Sized + NixPath>(
```
kubkon: Do we actually need the `NixPath` here at all? Since this is for our internal WASI use only, a standard `&str`/`String` should suffice, am I right? :-)

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,291 @@
...
+            Some(Ok(Entry(ent.assume_init())))
+        }
+    }
+}
+
+// impl<'d> Drop for Iter<'d> {
+//     fn drop(&mut self) {
+//         unsafe { libc::rewinddir((self.0).0.as_ptr()) }
+//     }
+// }
```
kubkon: Is there a reason why this is left out here commented out? Or could we safely remove it?

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,291 @@
...
+    CharacterDevice,
+    Directory,
+    BlockDevice,
+    File,
+    Symlink,
+    Socket,
+}
+
+impl Into<crate::hostcalls_impl::FileType> for Option<Type> {
+    fn into(self) -> crate::hostcalls_impl::FileType {
```
kubkon: Just a minor nitpick, but a `use crate::hostcalls_impl::FileType` and `impl Into<FileType>` would look neater here ;-)

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,291 @@
...
+        use crate::hostcalls_impl::FileType;
```
kubkon: Also, if we did a global `use crate::hostcalls_impl::FileType` as suggested in the comment above, we wouldn't need this one here.

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,291 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        unsafe {
+            // Note: POSIX specifies that portable applications should dynamically allocate a
+            // buffer with room for a `d_name` field of size `pathconf(..., _PC_NAME_MAX)` plus 1
+            // for the NUL byte. It doesn't look like the std library does this; it just uses
+            // fixed-sized buffers (and libc's dirent seems to be sized so this is appropriate).
+            // Probably fine here too then.
+            let mut ent = std::mem::MaybeUninit::<dirent>::uninit();
+            let mut result = ptr::null_mut();
+            if let Err(e) = Errno::result(readdir_r(
```
kubkon: Given that `readdir_r` is marked for deprecation, could we use `readdir` here instead? It would unify the implementation between Linux and BSD-style nixes (the latter uses `readdir` in our implementation of `fd_readdir` on BSD-style nixes).

src/sys/unix/dir.rs
kubkon: Also, correct me if I'm wrong but I guess `Dir::open` is obsolete for WASI applications since we'll only ever open a path relative to a file descriptor?

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,291 @@
...
+                (self.0).0.as_ptr(),
+                ent.as_mut_ptr(),
+                &mut result,
+            )) {
+                return Some(Err(e));
+            }
+            if result.is_null() {
+                return None;
+            }
+            assert_eq!(result, ent.as_mut_ptr());
```
kubkon: Hmm, could we somehow replace this `assert` with a `Result` instead? I'd really like our codebase to always (if possible) return an actual error (code) rather than panicking with a cryptic error message.

src/sys/unix/linux/hostcalls_impl.rs
```diff
@@ -29,75 +29,70 @@ pub(crate) fn path_rename(resolved_old: PathGet, resolved_new: PathGet) -> Resul
...
+    mut host_buf: &mut [u8],
+    cookie: host::__wasi_dircookie_t,
+) -> Result<usize> {
+    let vec = fd_readdir_impl(os_file, cookie)?;
+    debug!("Received dirents: {:?}", vec);
+    let mut used = 0;
+    for dirent in vec {
+        let dirent_raw = dirent.to_raw()?;
+        let offset = dirent_raw.len();
+        host_buf[0..offset].copy_from_slice(&dirent_raw);
```
kubkon: I think this will panic if `host_buf` has length smaller than `offset` ([copy_from_slice](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice)).

src/sys/unix/linux/hostcalls_impl.rs
```diff
@@ -29,75 +29,70 @@ pub(crate) fn path_rename(resolved_old: PathGet, resolved_new: PathGet) -> Resul
...
+// but there's BSD stuff remaining
+pub(crate) fn fd_readdir(
+    os_file: &mut OsFile,
+    mut host_buf: &mut [u8],
+    cookie: host::__wasi_dircookie_t,
+) -> Result<usize> {
+    let vec = fd_readdir_impl(os_file, cookie)?;
+    debug!("Received dirents: {:?}", vec);
+    let mut used = 0;
+    for dirent in vec {
```
kubkon: I might be way off base here, but as far as I can see, `vec` will always hold all dir entries from `cookie` onwards, am I correct? AFAIK, the current upstream implementation, on the other hand, should only ever iterate over the number of entries that can fit the provided `host_buf`. Could we also incorporate that here? @sunfishcode what do you think?

kubkon (changes requested):
Thanks for the PR! Looks good, just a couple of points I'd like to discuss.

kubkon:
> @kubkon in a subsequent PR, would you like to either (a) adapt the forked `Dir` to work on *BSD so that we can reuse the code between OSes (b) expose a function returning `Vec<Entry>` so that we can reuse just the part consisting of copying data around

@marmistrz After we address any remaining points and land this PR, I'll be happy to redo the BSD implementation in a way that re-uses as much of the functionality introduced in this PR as possible :-)

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,291 @@
...
+///    * implements `AsRawFd`, so it can be passed to `fstat`, `openat`, etc.
+///      The file descriptor continues to be owned by the `Dir`, so callers must not keep a `RawFd`
+///      after the `Dir` is dropped.
+///    * can be iterated through multiple times without closing and reopening the file
+///      descriptor. Each iteration rewinds when finished.
+///    * returns entries for `.` (current directory) and `..` (parent directory).
+///    * returns entries' names as a `CStr` (no allocation or conversion beyond whatever libc
+///      does).
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct Dir(ptr::NonNull<libc::DIR>);
```
kubkon: Could this be made `pub(crate)` instead of `pub`?

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,291 @@
...
+}
+
+impl Drop for Dir {
+    fn drop(&mut self) {
+        unsafe { libc::closedir(self.0.as_ptr()) };
+    }
+}
+
+#[derive(Debug, Eq, Hash, PartialEq)]
+pub struct Iter<'d>(&'d mut Dir);
```
kubkon: Like above, could this be made `pub(crate)` instead of `pub`?

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,291 @@
...
+//         unsafe { libc::rewinddir((self.0).0.as_ptr()) }
+//     }
+// }
+
+/// A directory entry, similar to `std::fs::DirEntry`.
+///
+/// Note that unlike the std version, this may represent the `.` or `..` entries.
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
+#[repr(transparent)]
+pub struct Entry(dirent);
```
kubkon: Like above, could this be made `pub(crate)` instead of `pub`?

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,291 @@
...
+
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
+pub enum Type {
```
kubkon: Like above, could this be made `pub(crate)` instead of `pub`?

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    #[cfg(target_os = "linux")]
+    pub fn seek_loc(&self) -> SeekLoc {
+        unsafe { SeekLoc::from_raw(self.0.d_off) }
+    }
+}
+
+#[cfg(not(target_os = "android"))]
+#[derive(Clone, Copy, Debug)]
+pub struct SeekLoc(libc::c_long);
```
kubkon: Like above, could this be made `pub(crate)` instead of `pub`?

src/sys/unix/linux/hostcalls_impl.rs
sunfishcode: Yes, that seems true. And in fact, if I'm reading this correctly, this implies there's quadradic behavior here if you're walking through a large directory with a fixed-size buffer doing `fd_readdir` to fill in one buffer at a time, because the number of `fd_readdir` calls is proportional to the size of the directory, and the number of entries each does is proportional to the size of the directory on average.

How hard would it be to teach `fd_readdir_impl` to serialize as it goes, so that it can know when it's reached the end of the buffer and stop? If it makes things easier, I'd say it's worth it to duplicate some code between platforms to help.

src/sys/unix/dir.rs
marmistrz: Probably yes. I wanted to introduce as little changes as possible, so that we can easily catch up with upstream if needed. Do you think it would be better to introduce a minimal module instead?

src/sys/unix/dir.rs
marmistrz: So that it's clear that this was present in the original code. Probably I should've added a comment.

src/sys/unix/dir.rs
marmistrz: This is a good argument for straying further away from the original nix implementation. Also see this related Rust issue: https://github.com/rust-lang/rust/issues/34668

src/sys/unix/linux/hostcalls_impl.rs
marmistrz: Good catch! We probably should return an iterator, not a `Vec`. I'll fix it.

src/sys/unix/dir.rs
marmistrz: This code was removed.

src/sys/unix/dir.rs
marmistrz: This code was removed.

src/sys/unix/dir.rs
marmistrz: This code was removed.

src/sys/unix/dir.rs
marmistrz: This code was removed.

src/sys/unix/dir.rs
marmistrz: Let's leave `readdir_r` for the time being, so that we can merge this PR and switch to `readdir` in a subsequent PR.

src/sys/unix/linux/hostcalls_impl.rs
marmistrz: This was changed to an iterator.

src/sys/unix/dir.rs
marmistrz: If this assertion fails, then `readdir_r` implementation doesn't adhere to the spec. I added a better panic message here.

src/sys/unix/linux/hostcalls_impl.rs
marmistrz: Good catch! Fixed.

marmistrz:
This is ready for another round of review.

src/hostcalls_impl/fs.rs
```diff
@@ -1046,3 +1048,43 @@ impl FileType {
...
+#[derive(Debug, Clone)]
+pub(crate) struct Dirent {
+    pub name: String,
+    pub ftype: FileType,
+    pub ino: u64,
+    pub cookie: wasi::__wasi_dircookie_t,
+}
+
+impl Dirent {
+    #![allow(unused)] // temporarily, until BSD catches up with this change
```
kubkon, line 1061: Hint, hint, get cracking Jakub! Just a joke, couldn't help myself ;-)

src/hostcalls_impl/fs.rs
```diff
@@ -1046,3 +1048,43 @@ impl FileType {
...
+    pub fn to_raw(&self) -> Result<Vec<u8>> {
```
kubkon: Hmm, I'm wondering here that perhaps `to_wasi_raw` would be a better name than `to_raw`? The latter kinda implies that it could be any sequence of bytes with some internal structure, whereas the former makes it clear that it's going to be a sequence of bytes that can be interpreted correctly on Wasm side. Additionally, a comment here explaining what this sequence of bytes is meant to contain would be great.

kubkon: Also, I'd consider making the entire method `unsafe` as, after all, we are casting the struct into a sequence of bytes which is indeed potentially unsafe.

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,214 @@
+// Based on src/dir.rs from nix
+#![allow(clippy::use_self)]
```
kubkon: Is this needed here? Perhaps it would be even better to have this in the main `lib.rs` manifest? A topic for another PR perhaps?

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,214 @@
...
+
+impl Dir {
+    /// Converts from a descriptor-based object, closing the descriptor on success or failure.
+    #[inline]
+    pub(crate) fn from<F: IntoRawFd>(fd: F) -> Result<Self> {
+        Dir::from_fd(fd.into_raw_fd())
+    }
+
+    /// Converts from a file descriptor, closing it on success or failure.
+    pub(crate) fn from_fd(fd: RawFd) -> Result<Self> {
```
kubkon: I'm wondering whether we actually want this exposed to the crate? I'm somewhat leaning towards making this method private to avoid confusion as to which one should be used: `from` or `from_fd`. Also, following the Rust's convention, we should, at the very least, make `from_fd` `unsafe`. I'm not sure whether `from` should also be `unsafe` though. Thoughts?

src/sys/unix/dir.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+pub(crate) struct IntoIter(Dir);
+impl Iterator for IntoIter {
+    type Item = Result<Entry>;
+    fn next(&mut self) -> Option<Self::Item> {
+        unsafe {
+            // Note: POSIX specifies that portable applications should dynamically allocate a
+            // buffer with room for a `d_name` field of size `pathconf(..., _PC_NAME_MAX)` plus 1
+            // for the NUL byte. It doesn't look like the std library does this; it just uses
```
kubkon, line 100: Would you mind adding a link here to the bit where this resides in the std library?

src/sys/unix/linux/hostcalls_impl.rs
```diff
@@ -67,75 +67,74 @@ pub(crate) fn path_rename(resolved_old: PathGet, resolved_new: PathGet) -> Resul
...
+    //     to opendir() or rewinddir(), whether a subsequent call to readdir() returns an entry
+    //     for that file is unspecified.
+    if cookie == wasi::__WASI_DIRCOOKIE_START {
+        trace!("     | fd_readdir: doing rewinddir");
+        dir.rewind();
     } else {
-        // If cookie set to __WASI_DIRCOOKIE_START, rewind the dir ptr
-        // to the start of the stream.
-        unsafe { rewinddir(dir) };
+        trace!("     | fd_readdir: doing seekdir to {}", cookie);
```
kubkon, line 96: Love the `trace` calls here, thanks!

src/sys/unix/mod.rs
```diff
@@ -2,6 +2,9 @@ pub(crate) mod fdentry_impl;
 pub(crate) mod host_impl;
 pub(crate) mod hostcalls_impl;
 
+#[allow(trivial_numeric_casts)]
```
kubkon: Do we actually need this attribute here, or is a leftover perhaps?

src/sys/windows/hostcalls_impl/fs.rs
```diff
@@ -166,12 +168,117 @@ pub(crate) fn path_open(
...
+) -> Result<Dirent> {
+    let path = path.as_ref();
+    trace!("dirent_from_path: opening {}", path.to_string_lossy());
+
+    // To open a directory on Windows, FILE_FLAG_BACKUP_SEMANTICS flag must be used
+    let file = OpenOptions::new()
+        .custom_flags(Flags::FILE_FLAG_BACKUP_SEMANTICS.bits())
+        .read(true)
+        .open(path)?;
+    let ty = file.metadata()?.file_type();
```
kubkon, line 184: Should we worry here that `std::fs::metadata` will only allow us to differentiate between a file, dir and symlink on Windows? When determining the type and access rights for a `Handle` in `fdentry_impl`, we use [get_file_type] to discern into more types than a file, dir, and symlink, but I'm wondering whether there is any point in doing that here. I'm not that familiar with Windows unfortunately, but perhaps @peterhuene could you offer a word of advice on this?

[get_file_type]: https://github.com/CraneStation/wasi-common/blob/master/winx/src/file.rs#L49

src/sys/unix/linux/hostcalls_impl.rs
kubkon: The applied changes look good to me, but I'll hand it over to @sunfishcode to see what he thinks.

src/sys/unix/mod.rs
marmistrz: Yes and no. I changed those to into conversions, so that they fail on bleeding-edge platforms where they could be lossy.

src/sys/unix/dir.rs
marmistrz: A leftover from the previous implementation, but I definitely agree.  Fixed now.

src/sys/unix/dir.rs
marmistrz: I added it back in the times when I didn't want to stray far from the original implementation. It's fixed now.

src/sys/windows/hostcalls_impl/fs.rs
marmistrz: In the `filestat` family we're not supporting those various types either. WASI doesn't support FIFOs and pipes for the time being. Let's open an issue for now and take care of it later

I also think that what `get_file_type` should ideally be a part of [`FileTypeExt`  in libstd](https://doc.rust-lang.org/std/os/windows/fs/trait.FileTypeExt.html)

src/sys/unix/dir.rs
marmistrz: Doine.

src/hostcalls_impl/fs.rs
marmistrz: > `to_wasi_raw`

That's a good idea. An alternative could possibly be something `serialize_wasi_raw`, but `to_wasi_raw` sound perfectly fine. I also added some docs.

IMO this is fine, just as it's fine to create a raw pointer from a reference, but it's unsafe to try to dereference it. In other words, trying to interpret this sequence of bytes as any kind of object would definitely be unsafe but serializing the object is perfectly safe.

marmistrz:
I addressed all the comments from the latest review.

kubkon (approved):
Great job!

kubkon:
I reckon we're good to go with this one as-is. If we find any areas that need fixing/improving, I suggest we address those in subsequent PRs.

PR #117, opened 2019-10-07
Use cvt from the cvt crate in winx

marmistrz:
`cvt` was copied over from Rust libstd in https://github.com/CraneStation/wasi-common/pull/42, but I've created a separate crate for it, so as to be able to reuse it across crates.

### DIFF
--- a/winx/Cargo.toml
+++ b/winx/Cargo.toml
@@ -9,3 +9,4 @@ description = "Windows API helper library"
 [dependencies]
 winapi = { version = "0.3", features = ["std", "errhandlingapi", "handleapi", "processthreadsapi", "securitybaseapi", "winbase", "winerror", "ws2def", "fileapi", "aclapi" ] }
 bitflags = "1.0"
+cvt = "0.1"
--- a/winx/src/file.rs
+++ b/winx/src/file.rs
@@ -1,5 +1,6 @@
 #![allow(non_camel_case_types)]
 use crate::{winerror, Result};
+use cvt::cvt;
 use std::ffi::{c_void, OsString};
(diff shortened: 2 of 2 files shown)


pchickey:
Hi, I'm pretty ignorant of all things windows programming, and unfortunately I've read this PR and the entirety of the `cvt` crate and I still don't know what cvt stands for or why it is used. Can you add two sentences of docs to the cvt crate explaining what its about, and why we're using a copy from libstd as a crate?

programmerjake:
It would appear as the `cvt` function handles translating return values from Win32/Unix/VxWorks APIs to `std::io::Result`.

programmerjake:
`cvt_r` handles retrying the API call when it returns `ErrorKind::Interrupted`

marmistrz:
I also have no idea why `cvt` was named that way :)

What @programmerjake said is exactly what `cvt` does. Different systems (especially Windows vs Unix) do it in a slightly different way (0 signifies success on Unix but failure on Windows), so `cvt` is there to reduce the mental bookkeeping and make it easier to handle errors from syscalls.

I've just realized that the crate has no crate-level documentation, but there's some description in the [git repo](https://github.com/marmistrz/cvt). Is that one okay or should be a little more verbose in the crate description?

We're using a copy from libstd, because `libstd::sys` is not public. :) Should I also mention this in the docs?

pchickey:
That's very helpful. thank you! Those details would be very helpful in the crate docs.

marmistrz:
I have updated the crate docs and published a new version of the crate.

kubkon (approved):
LGTM, thanks!

PR #118, opened 2019-10-08
moved preopen_path field out of FdEntry

ibaryshnikov:
This PR introduces a new struct
```rust
struct FileDetails {
    file: OsFile,
    preopen_path: Option<PathBuf>,
}
```
Motivation: as `preopen_path` doesn't make sense with Stdin, Stdout, Stderr, and in the future with Socket file descriptors, I suggest moving it inside a `Descriptor::OsFile` enum variant.

Drawbacks: 
increased verbosity in some cases:
```
-fe.preopen_path = Some(guest_path);
+fe.fd_object.descriptor.as_file_details_mut()?.preopen_path = Some(guest_path);
```

I'd like to hear any feedback about this change.

P.S.
Thanks @sunfishcode for your mentorship!

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -115,7 +115,7 @@ impl WasiCtxBuilder {
                 preopen_fd = preopen_fd.checked_add(1).ok_or(Error::ENFILE)?;
             }
             let mut fe = FdEntry::from(dir)?;
-            fe.preopen_path = Some(guest_path);
+            fe.fd_object.descriptor.as_file_details_mut()?.preopen_path = Some(guest_path);
             self.fds.insert(preopen_fd, fe);
             preopen_fd += 1;
         }
--- a/src/fdentry.rs
+++ b/src/fdentry.rs
@@ -5,8 +5,14 @@ use std::path::PathBuf;
 use std::{fs, io};
 
 #[derive(Debug)]
+pub struct FileDetails {
+    pub(crate) file: OsFile,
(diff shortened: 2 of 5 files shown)


sunfishcode:
While preopen paths are currently limited to directories, in the future it's likely that we'll support preopen for [files too](https://github.com/WebAssembly/WASI/issues/91). In that case, I could imagine wanting to remap an application requesting to read from "/some/path" to actually read from the host process' stdin instead, or also a socket. I haven't though about this too deeply yet though. What do you think?

programmerjake:
> While preopen paths are currently limited to directories, in the future it's likely that we'll support preopen for [files too](https://github.com/WebAssembly/WASI/issues/91). In that case, I could imagine wanting to remap an application requesting to read from "/some/path" to actually read from the host process' stdin instead, or also a socket. I haven't though about this too deeply yet though. What do you think?

Seems worth doing, though most programs don't expect reading files to block (pipes/sockets in the filesystem are uncommon on Windows).

ibaryshnikov:
It means there's still a need of `preopen_path` inside `FdEntry`, right?
Instead, to support preopen files and sockets, there's a need to update this place?
https://github.com/CraneStation/wasi-common/blob/968fb19f17ebbcb982939d77cdce7d2dd41964c9/src/ctx.rs#L110-L118

sunfishcode:
Yes; actually supporting other kinds of preopens can wait until WASI itself adds them. For now, we can just keep this possibility in mind in the code.

PR #119, opened 2019-10-09
Implement clock_time_get on Windows

marmistrz:
We still need to have https://github.com/tailhook/cpu-time/pull/2 merged to avoid panics in case of errors. If you think it's unnecessary, let me know. For the time being, I'm using my fork which return proper `Result`s.


kubkon:
In the future, I'd be good if we could ensure that whenever making a PR we only change bits that are directly relevant. For instance, this PR fixes several Windows warnings, which is good, but I'd much prefer if we did fixes like this in a separate PR (I'm thinking here of changes introduced in the f8a90c9 commit). Anyhow, it's just a pointer for the future that I hope we all can follow (I'm especially thinking here of myself as I'm notorious for introducing tiny changes in PRs that are relevant). ;-)

kubkon:
In terms of using `cpu-time` crate. I like the idea, I'm not sure we want to go this route though for two reasons: 1) I'm not sure how likely this crate is to be maintained in the future which can become a headache down the line for us, and 2) if the author is reluctant to actually throw errors instead of panics as pointed out in tailhook/cpu-time#2, we might want to consider implementing our own syscall wrappers. TBH it doesn't seem like a lot of work to simply do it ourselves, but I'm open for discussion on this front. @sunfishcode thoughts?

marmistrz:
Well, if the author ceases to maintain the crate or refuses to throw errors, we can always fork it.

sunfishcode:
@kubkon Agreed; I am also guilty of sometimes adding unrelated cleanups along with a PR :-}, but it can impede reviews. So if you or anyone is reviewing a patch which is awkward, we should feel free to request that it be split up into separate changes.

At a quick glance, it looks like the `cpu-time` maintainer is responding, in which case we may be ok. But if we hit trouble, maintaining our own wrappers is within reason as well.

marmistrz:
Now we no longer depend on the git version of any crates, the relevant change was successfully upstreamed.

src/sys/windows/hostcalls_impl/misc.rs
```diff
@@ -1,9 +1,18 @@
...
-use crate::{host, wasm32, Result};
+use crate::{host, wasm32, Error, Result};
+use cpu_time::{ProcessTime, ThreadTime};
+use lazy_static::lazy_static;
+use std::convert::TryInto;
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
+
+lazy_static! {
+    static ref START_MONOTONIC: Instant = Instant::now();
+}
```
kubkon, line 15: This results mainly from my lack of knowledge and experience with `std::time::Instant` but is it thread-safe? Or should we wrap in some sort of a guard here?

kubkon (approved):
LGTM! Just one tiny question.

src/sys/windows/hostcalls_impl/misc.rs
marmistrz: I'm not really sure what you mean. This variable is initialized once and is read-only later on. There can be no race condition if a variable is read-only.

src/sys/windows/hostcalls_impl/misc.rs
```diff
@@ -1,9 +1,18 @@
...
 
 use wasi_common_cbindgen::wasi_common_cbindgen;
```
kubkon:
```suggestion
```

kubkon: Looks like a leftover to me that we can recycle while we're doing other Win cleanups in this PR.

src/sys/windows/hostcalls_impl/misc.rs
kubkon: That's exactly the answer I was looking for. I just wasn't sure/didn't remember whether `elapsed` was somehow mutating the state or not, but I see that it's not. So just wanted to make sure that it's not a potential thread-safety issue.

kubkon (commented):
Could you please rebase to latest `master`?

PR #124, opened 2019-10-11
Try to somehow implement clock_res_get on Windows. 

marmistrz:
I don't like this implementation, and I'm not sure if we should merge it, but I think it's the best we can do without overly complicating stuff. We depend on the implementation details on `std::time::{SystemTime, Instant}` and `cpu-time` (on the underlying Windows syscalls).

While we could avoid it for `ProcessTime` and `ThreadTime` by adding an associated function returning the timer resolution, possibly the same for `Instant` (and wait forever until it's stable), but the problems for `SystemTime` are much more fundamental. We may have to reconsider this part of WASI API.

Depends on #117 and the implementation details of #119.

### DIFF
--- a/src/sys/windows/hostcalls_impl/misc.rs
+++ b/src/sys/windows/hostcalls_impl/misc.rs
@@ -52,7 +52,7 @@ pub(crate) fn clock_res_get(clock_id: wasi::__wasi_clockid_t) -> Result<wasi::__
         // [4] https://www.codeproject.com/Tips/1011902/High-Resolution-Time-For-Windows
         // [5] https://stackoverflow.com/questions/7685762/windows-7-timing-functions-how-to-use-getsystemtimeadjustment-correctly
         // [6] https://bugs.python.org/issue19007
-        wasi::__WASI_CLOCK_REALTIME => 55000,
+        wasi::__WASI_CLOCK_REALTIME => 55_000_000,
         // std::time::Instant uses QueryPerformanceCounter & QueryPerformanceFrequency internally
         wasi::__WASI_CLOCK_MONOTONIC => get_perf_counter_resolution()?.as_nanos().try_into()?,
         // The best we can do is to hardcode the value from the docs.
--- a/src/sys/windows/hostcalls_impl/misc.rs
+++ b/src/sys/windows/hostcalls_impl/misc.rs
@@ -5,7 +5,6 @@ use crate::helpers::systemtime_to_timestamp;
 use crate::hostcalls_impl::{ClockEventData, FdEventData};
 use crate::memory::*;
 use crate::sys::host_impl;
-use crate::sys::host_impl;
 use crate::{wasi, wasi32, Error, Result};
 use cpu_time::{ProcessTime, ThreadTime};
 use lazy_static::lazy_static;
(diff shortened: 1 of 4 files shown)


src/sys/windows/hostcalls_impl/misc.rs
```diff
@@ -1,18 +1,80 @@
...
+        // For instance, the intervals return by the syscall are:
+        // * [1, 65536] on Wine
+        // * [1, 1000000] on Windows 10, which is up to (sic) 1000 seconds.
+        //
+        // It's possible to manually set the timer resolution, but this sounds like something which should
+        // only be done temporarily. [5]
+        //
+        // Alternatively, we could possibly use GetSystemTimePreciseAsFileTime in clock_time_get, but
+        // this syscall is only available starting from Windows 8.
+        // (we could possibly emulate it on earlier versions of Windows, see [4])
```
kubkon, line 42: Should we even worry about Windows versions lower than Windows 8? After all, Microsoft is meant to end the extended support (mainstream is already finished for Windows 7 and 8.1) for Windows 7 by January 2020 (see [this link]).

[this link]: https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet

programmerjake: I think having a fallback for Windows 7 is a good idea, since there are lots of people who can't/won't upgrade (for example, if they have custom drivers that they can't port to newer versions of Windows).

kubkon: Fair enough. Still, it would be good to leverage the new syscalls whenever possible. @marmistrz do you think it would be possible to have an implementation that tries to use `GetSystemTimePreciseAsFileTime`, and if that fails, it reverts back to the current implementation? I'm thinking here a mechanism similar to the one we've implemented in @alexcrichton's [filetime] crate.

[filetime]: https://github.com/alexcrichton/filetime/blob/master/src/unix/macos.rs#L94

marmistrz: We should definitely support Windows 7. According to netmarketshare.com, Windows 7 still has about 35% market share. It's unclear whether or not we should support Windows XP, which has about 2.5% market share (which is incidentally higher than Linux's)

We could, but I'm not sure if we should. Even OpenJDK refused to switch to `GetSystemTimePreciseAsFileTime` this February because of the following issue. See [this link](https://bugs.openjdk.java.net/browse/JDK-8180466) for their rationale.

sunfishcode:
Unforunately, `SystemTime` is the most important one. My sense is that the approach in this PR is acceptable here, if that's our best option.

`ProcessTime` and `ThreadTime` may end up getting removed from WASI, since we generally don't want to require that WASI programs be one-to-one with host processes or host threads, and I don't know of an efficient way to implement them on popular OS's otherwise.

marmistrz:
I caught up with latest master.

src/sys/windows/hostcalls_impl/misc.rs
kubkon: SGTM!

marmistrz:
I'll probably need to open an issue in WebAssembly/WASI about these issues with implementation. They may make us reconsider the API spec.

winx/src/time.rs
```diff
@@ -0,0 +1,10 @@
+use cvt::cvt;
+use winapi::um::{profileapi::QueryPerformanceFrequency, winnt::LARGE_INTEGER};
+
+pub fn perf_counter_frequency() -> std::io::Result<u64> {
+    unsafe {
+        let mut frequency: LARGE_INTEGER = std::mem::zeroed();
+        cvt(QueryPerformanceFrequency(&mut frequency))?;
```
kubkon, line 7: Here's a thought. Given that the MSDN docs on `QueryPerformanceFrequency` state that

> The frequency of the performance counter is fixed at system boot and is consistent across all processors. Therefore, the frequency need only be queried upon application initialization, and the result can be cached.

(here's the [link](https://docs.microsoft.com/en-us/windows/win32/api/profileapi/nf-profileapi-queryperformancefrequency)) would it be prudent for us to adhere to this and actually cache the result of the function, and then simply re-use the result rather than perform a syscall to essentially read out the same value over and over? @peterhuene thoughts?

kubkon:
> I'll probably need to open an issue in WebAssembly/WASI about these issues with implementation. They may make us reconsider the API spec.

I'm not counting that out. However, before we do that, perhaps we should consult @peterhuene first? @peterhuene would you have any thoughts on this PR and the discussion in general?

src/sys/windows/hostcalls_impl/misc.rs
```diff
@@ -61,3 +107,10 @@ fn get_proc_cputime() -> Result<Duration> {
 fn get_thread_cputime() -> Result<Duration> {
     Ok(ThreadTime::try_now()?.as_duration())
 }
+
+fn get_perf_counter_resolution() -> Result<Duration> {
```
peterhuene: Should this return a `u64` rather than `Duration`?  It seems that the resolution is always expressed as the number of nanoseconds and we do a conversion back to nanoseconds immediately upon calling this function.

winx/src/time.rs
peterhuene: I think caching it would be a good idea.

src/sys/windows/hostcalls_impl/misc.rs
```diff
@@ -15,8 +15,54 @@ lazy_static! {
...
+        // See also this discussion about the use of GetSystemTimePreciseAsFileTime in Python stdlib,
+        // which in particular contains some resolution benchmarks.
+        //
+        // [1] https://devblogs.microsoft.com/oldnewthing/20170921-00/?p=97057
+        // [2] http://www.windowstimestamp.com/description
+        // [3] https://docs.microsoft.com/en-us/windows/win32/sysinfo/acquiring-high-resolution-time-stamps?redirectedfrom=MSDN
+        // [4] https://www.codeproject.com/Tips/1011902/High-Resolution-Time-For-Windows
+        // [5] https://stackoverflow.com/questions/7685762/windows-7-timing-functions-how-to-use-getsystemtimeadjustment-correctly
+        // [6] https://bugs.python.org/issue19007
+        wasi::__WASI_CLOCK_REALTIME => 55000,
```
peterhuene: Should this be `55_000_000` if we're representing 55ms (the potential upper-bound of the resolution for `GetSystemTimeAsFileTime`) in nanoseconds?

It might be worth considering the fact that Windows 7 is EOL in a few short months and switch to using the precise API if we think users will want sub-microsecond resolution of the realtime clock, potentially falling back to `std::time::SystemTime` if the precise API is not available.

If we do switch to `GetSystemTimePreciseAsFileTime`, I think we could get away with assuming the "precise" system time resolution is the same as the monotomic resolution since it seems to use QPC for counting the elapsed time since the last timer interrupt.  But if we're not confortable with that, then I think a microsecond (i.e. `1000`) would be a relatively safe upper-bound, much like we're using 55ms here for the imprecise API.

peterhuene (changes requested):
This looks great and you did some really in-depth investigation into how this could be well-implemented on Windows.  Thanks for that!

Just some general feedback and a requested clarification on what the resolution value for the realtime clock is representing.

src/sys/windows/hostcalls_impl/misc.rs
marmistrz: Please see discussion in @kubkon's review from 16th Oct. Feel free to unresolve it if you think it was resolved prematurely.

peterhuene: I'm 👍 with not using a higher precision real time clock for now (it's something that can always be changed later if users really need it).  However, can you verify for me that the value of `55000` is correct?  Is this 55 μs or 55 ms?

marmistrz: Yes, this should definitely be 55ms, thanks for noticing that!
Maybe I should rather use `Duration::from_millis(55).as_nanos().try_into().unwrap()`, but it's getting messy with conversion (or use `as`)

winx/src/time.rs
marmistrz: This is definitely a good idea and is fixed now, but requires a somewhat lossy error conversion (but I wouldn't worry about it too much). We're hit by https://github.com/rust-lang/rust/issues/24135.

peterhuene: I wonder if we should only cache it on success rather than caching the error as well.

Additionally, it looks like we could also just panic if it fails:

```
On systems that run Windows XP or later, the function will always succeed and will thus never return zero.
```

marmistrz: I think it's perfectly fine to panic on systems earlier than Windows XP.

src/sys/windows/hostcalls_impl/misc.rs
marmistrz: The way I see it, `Instant`s from the stdlib are treated similar to points in an affine space having `Duration`s as its tangent space (aka free vectors) - you can measure the distance between two `Instant`s and also add a `Duration` to an `Instant`. (this is some fuzzy intuition, not a precise statement). Resolution also sounds like a free vector in here, so `Duration` seems semantically fine.

On the other hand, packing and unpacking the integer does no good for code readability, so I think it's a perfectly valid choice to return a `u64` and I'm doing so in the latest revision of the PR.

peterhuene (approved):

sunfishcode (approved):

kubkon (approved):
Great job, thanks!

PR #125, opened 2019-10-15
Fixes `path_symlink_trailing_slashes` test case

kubkon:
This PR:
* adds a couple `log::debug!` macro calls in and around `path_get`
  for easier future debugging
* changes impl of `path_symlink` hostcall to actually *require*
  the final component (matching the impl of WASI in C)
* ignores the error `__WASI_ENOTDIR` in `path_get`'s `readlinkat` call
  which is not meant to be an error at this stage (i.e., this
  potentially erroneous condition *will be* handled later, in
  one of the layers above)
* makes `path_symlink` host-specific (Linux and BSD-like nixes
  now have their own differing implementations)
* on BSD-like nixes, when `ENOTDIR` is returned from `symlinkat`
  it checks whether the target path contains a trailing slash,
  strips it, and then checks if the target path without the trailing
  slash exists; if yes, then converts the error code to `EEXIST` to
  match Linux/POSIX spec

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -170,15 +170,8 @@ mod wasm_tests {
     cfg_if::cfg_if! {
         if #[cfg(not(windows))] {
             /// Ignore tests that aren't supported yet.
-            fn ignore(testsuite: &str, name: &str) -> bool {
-                if testsuite == "misc_testsuite" {
-                    match name {
-                        "path_symlink_trailing_slashes" => true,
-                        _ => false,
-                    }
-                } else {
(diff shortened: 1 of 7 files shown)


kubkon:
@sunfishcode I hope you don't mind if I'm gonna go ahead and merge this one so that we get all tests covered under Linux and MacOS hosts (and hopefully a couple BSDs). If you find anything amiss, open an issue and lemme know please! :-)

PR #131, opened 2019-10-16
Fixes CraneStation/wasmtime#396

kubkon:
This commit fixes an issue with incorrect handling of /dev/(u)random
on Linux. It turns out that `nix::unistd::isatty` call handled only
the POSIX spec case where `ENOTTY` is returned in case the passed
in file descriptor is OK but not a TTY, whereas on Linux this is not
always the case. On Linux, it can be the case that `EINVAL` is returned
instead and this case AFAIK is not handled by the `nix` crate. This
commit fixes this by using `libc::isatty` syscall directly and checking
the return values.

### DIFF
--- a/src/sys/unix/bsd/mod.rs
+++ b/src/sys/unix/bsd/mod.rs
@@ -1,2 +1,19 @@
 pub(crate) mod hostcalls_impl;
 pub(crate) mod osfile;
+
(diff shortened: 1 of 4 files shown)


sunfishcode (commented):
Great work tracking that down!

sunfishcode (approved):

PR #133, opened 2019-10-18
Ensure that each crate has its own copy of the LICENSE file.

sunfishcode:
On crates.io, each crate is distributed separately, so ensure that each
crate has a license file.


kubkon:
LGTM!

PR #134, opened 2019-10-18
Fix nondeterminism in the preopen order.

sunfishcode:
We iterate over the preopens to present them to the WASI program, so
storing them in a `HashMap` means this order is nondeterministic. Switch
to a `Vec` of tuples instead. This means we don't eliminate duplicates,
but they should be rare.


kubkon (approved):
LGTM! Did this have any noticeable effect anywhere? I'm asking just out of sheer curiosity cause using `HashMap` in place of a `Vec` would obviously lead to nondeterminism. :-)

sunfishcode:
I didn't see any noticeable effects; I just saw it when I was debugging some unrelated issues and saw things changing order from run to run.

PR #137, opened 2019-10-21
Refactor poll_oneoff on *nix

### LINKED ISSUE #440: poll_oneoff with wasi::EVENTTYPE_FD_READ resulting in EBADF
In the following code compiled using `rust 1.40 nightly` when passing `fd` from preopened file I get error code 8 as output event from poll_oneoff. I guess it should not be bad fd since reading from file succeeds?
```
#![feature(wasi_ext)]
use ::wasi::wasi_unstable as wasi;
use std::os::wasi::io::AsRawFd;
use std::io::prelude::*;

fn main() {
    let mut input = std::fs::File::open("input.txt").expect("file input.txt");
    let mut buf: String = String::new();
    println!("{:?}", input.read_to_string(&mut buf));

    println!("File descriptor: {}", input.as_raw_fd());

    let fd_readwrite = wasi::raw::__wasi_subscription_u_fd_readwrite_t {
        fd: input.as_raw_fd()
    };
    let in_ = [wasi::Subscription {
        userdata: 1,
        type_: wasi::EVENTTYPE_FD_READ,
        u: wasi::raw::__wasi_subscription_u { fd_readwrite },
    }];
    let (res, event) = unsafe {
        let mut out: [wasi::Event; 1] = std::mem::zeroed();
        let res = wasi::poll_oneoff(&in_, &mut out);
        (res, out[0])
    };

    println!("Res = {:?}", res);
    assert_eq!(event.userdata, 1, "event.userdata");
    assert_eq!(event.error, 0, "event.error");
    assert_eq!(event.type_, wasi::EVENTTYPE_FD_READ, "event.type_");
}
```

Following repo has required setup and `test.sh` script compiling and running through wasmtime https://github.com/dunnock/poll_oneoff_tests:
```
poll_oneoff_tests % ./test.sh
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
Ok(2)
File descriptor: 4
Res = Ok(1)
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `8`,
 right: `0`: event.error', src/main.rs:97:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
error: failed to process main module `target/wasm32-wasi/debug/poll_oneoff_tests.wasm`
    caused by: Instantiation error: Trap occurred while invoking start function: wasm trap: unreachable, source location: @19834
```

with `RUST_LOG=trace`:
```
...
TRACE wasmtime_wasi::syscalls              > path_open(dirfd=3, dirflags=1, path=0x110128, path_len=9, oflags=0x0, fs_rights_base=0xf87febe, fs_rights_inheriting=0xf87febe, fs_flags=0x0, fd=0xffa88)
 TRACE wasi_common::hostcalls_impl::fs      > path_open(dirfd=3, dirflags=1, path_ptr=0x110128, path_len=9, oflags=0x0, fs_rights_base=0xf87febe, fs_rights_inheriting=0xf87febe, fs_flags=0x0, fd_out_ptr=0xffa88)
 TRACE wasi_common::hostcalls_impl::fs      >      | (path_ptr,path_len)='input.txt'
 DEBUG wasi_common::hostcalls_impl::fs_helpers > path_get cur_path = "input.txt"
 DEBUG wasi_common::hostcalls_impl::fs_helpers > path_get path_stack = []
 DEBUG wasi_common::sys::unix::hostcalls_impl::fs_helpers > path_get readlinkat path = "input.txt"
 DEBUG wasi_common::sys::unix::hostcalls_impl::fs         > path_open resolved = PathGet { dirfd: File { fd: 7, path: "/Users/max/src/wasm/poll_oneoff_tests/input", read: true, write: false }, path: "input.txt" }
 DEBUG wasi_common::sys::unix::hostcalls_impl::fs         > path_open oflags = O_NOFOLLOW
 DEBUG wasi_common::sys::unix::hostcalls_impl::fs         > path_open new_fd = 8
 DEBUG wasi_common::sys::unix::fdentry_impl               > Fd 8 is a file
 DEBUG wasi_common::sys::unix::fdentry_impl               > Fd 8 is a file
 TRACE wasi_common::hostcalls_impl::fs                    >      | *fd=4
 TRACE wasi_common::hostcalls                             >     -> errno=__WASI_ESUCCESS
 TRACE wasmtime_wasi::syscalls                            > fd_read(fd=4, iovs=0xff9e0, iovs_len=1, nread=0xff9ec)
 TRACE wasi_common::hostcalls_impl::fs                    > fd_read(fd=4, iovs_ptr=0xff9e0, iovs_len=1, nread=0xff9ec)
 TRACE wasi_common::hostcalls_impl::fs                    >      | *nread=2
 TRACE wasi_common::hostcalls                             >     -> errno=__WASI_ESUCCESS
 TRACE wasmtime_wasi::syscalls                            > fd_read(fd=4, iovs=0xff9e0, iovs_len=1, nread=0xff9ec)
 TRACE wasi_common::hostcalls_impl::fs                    > fd_read(fd=4, iovs_ptr=0xff9e0, iovs_len=1, nread=0xff9ec)
 TRACE wasi_common::hostcalls_impl::fs                    >      | *nread=0
 TRACE wasi_common::hostcalls                             >     -> errno=__WASI_ESUCCESS
 TRACE wasmtime_wasi::syscalls                            > fd_write(fd=1, iovs=0xff9e0, iovs_len=1, nwritten=0xff9e8)
 TRACE wasi_common::hostcalls_impl::fs                    > fd_write(fd=1, iovs_ptr=0xff9e0, iovs_len=1, nwritten=0xff9e8)
Ok(2)
 TRACE wasi_common::hostcalls_impl::fs                    >      | *nwritten=6
 TRACE wasi_common::hostcalls                             >     -> errno=__WASI_ESUCCESS
 TRACE wasmtime_wasi::syscalls                            > fd_write(fd=1, iovs=0xff9e0, iovs_len=1, nwritten=0xff9e8)
 TRACE wasi_common::hostcalls_impl::fs                    > fd_write(fd=1, iovs_ptr=0xff9e0, iovs_len=1, nwritten=0xff9e8)
File descriptor: 4
 TRACE wasi_common::hostcalls_impl::fs                    >      | *nwritten=19
 TRACE wasi_common::hostcalls                             >     -> errno=__WASI_ESUCCESS
 TRACE wasmtime_wasi::syscalls                            > poll_oneoff(in=0xffc08, out=0xffcf8, nsubscriptions=1, nevents=0xffaf0)
 TRACE wasi_common::hostcalls_impl::misc                  > poll_oneoff(input=0xffc08, output=0xffcf8, nsubscriptions=1, nevents=0xffaf0)
 TRACE wasi_common::hostcalls_impl::misc                  >      | *nevents=1
 TRACE wasi_common::hostcalls                             >     -> errno=__WASI_ESUCCESS
 TRACE wasmtime_wasi::syscalls                            > fd_write(fd=1, iovs=0xff9e0, iovs_len=1, nwritten=0xff9e8)
 TRACE wasi_common::hostcalls_impl::fs                    > fd_write(fd=1, iovs_ptr=0xff9e0, iovs_len=1, nwritten=0xff9e8)
Res = Ok(1)
...
```

kubkon:
This PR introduces a couple of changes/fixes:
* it annotates `log::debug!` messages with "host" to differentiate
  between file descriptors stored on the WASI side (aka the wrappers)
  and those managed by the host (aka the wrapped)
* it fixes CraneStation/wasmtime#440, i.e., incorrect passing of
  file descriptor to `poll_oneoff` where currently errenously we
  pass in the wrapper instead of the wrapped value
* it adds a couple more `log::debug!` macros calls for easier future
  debugging
* it refactors `poll_oneoff` for easier maintenance in the future (hopefully!)

### DIFF
--- a/src/hostcalls_impl/misc.rs
+++ b/src/hostcalls_impl/misc.rs
@@ -183,7 +183,7 @@ pub(crate) fn clock_time_get(
 }
 
 pub(crate) fn poll_oneoff(
-    _wasi_ctx: &WasiCtx,
+    wasi_ctx: &WasiCtx,
     memory: &mut [u8],
     input: wasm32::uintptr_t,
     output: wasm32::uintptr_t,
@@ -207,7 +207,7 @@ pub(crate) fn poll_oneoff(
     let input_slice = dec_slice_of::<wasm32::__wasi_subscription_t>(memory, input, nsubscriptions)?;
     let input: Vec<_> = input_slice.iter().map(dec_subscription).collect();
     let output_slice = dec_slice_of_mut::<wasm32::__wasi_event_t>(memory, output, nsubscriptions)?;
-    let events_count = hostcalls_impl::poll_oneoff(input, output_slice)?;
+    let events_count = hostcalls_impl::poll_oneoff(wasi_ctx, input, output_slice)?;
 
     trace!("     | *nevents={:?}", events_count);
 
--- a/src/hostcalls_impl/misc.rs
+++ b/src/hostcalls_impl/misc.rs
@@ -205,11 +205,8 @@ pub(crate) fn poll_oneoff(
     enc_pointee(memory, nevents, 0)?;
 
     let input_slice = dec_slice_of::<wasm32::__wasi_subscription_t>(memory, input, nsubscriptions)?;
-    let input: Vec<_> = input_slice
-        .iter()
-        .map(dec_subscription)
-        .collect::<Result<Vec<_>>>()?;
-    let events = hostcalls_impl::poll_oneoff(wasi_ctx, input)?;
+
+    let events = hostcalls_impl::poll_oneoff(wasi_ctx, input_slice.iter().map(dec_subscription))?;
 
     let output_slice = dec_slice_of_mut::<wasm32::__wasi_event_t>(memory, output, nsubscriptions)?;
     let mut events_count = 0;
--- a/src/hostcalls_impl/misc.rs
+++ b/src/hostcalls_impl/misc.rs
@@ -205,9 +205,18 @@ pub(crate) fn poll_oneoff(
     enc_pointee(memory, nevents, 0)?;
 
     let input_slice = dec_slice_of::<wasm32::__wasi_subscription_t>(memory, input, nsubscriptions)?;
-    let input: Vec<_> = input_slice.iter().map(dec_subscription).collect();
+    let input: Vec<_> = input_slice
+        .iter()
+        .map(dec_subscription)
+        .collect::<Result<Vec<_>>>()?;
+    let events = hostcalls_impl::poll_oneoff(wasi_ctx, input)?;
+
     let output_slice = dec_slice_of_mut::<wasm32::__wasi_event_t>(memory, output, nsubscriptions)?;
-    let events_count = hostcalls_impl::poll_oneoff(wasi_ctx, input, output_slice)?;
+    let mut events_count = 0;
+    for (i, event) in events.into_iter().enumerate() {
+        output_slice[i] = enc_event(event);
+        events_count += 1;
+    }
 
     trace!("     | *nevents={:?}", events_count);
 
--- a/src/hostcalls_impl/misc.rs
+++ b/src/hostcalls_impl/misc.rs
@@ -222,10 +222,10 @@ pub(crate) fn poll_oneoff(
         match event.type_ {
             host::__WASI_EVENTTYPE_CLOCK => {
                 let clock = unsafe { event.u.clock };
-                let delay = wasi_clock_to_relative_ns_delay(clock)? / 1_000_000; // in milliseconds!
+                let delay = wasi_clock_to_relative_ns_delay(clock)?;
 
                 log::debug!("poll_oneoff event.u.clock = {:?}", clock);
-                log::debug!("poll_oneoff delay = {:?}ms", delay);
+                log::debug!("poll_oneoff delay = {:?}ns", delay);
 
                 let current_event = ClockEventData {
                     delay,
@@ -299,7 +299,7 @@ fn wasi_clock_to_relative_ns_delay(
 
 #[derive(Debug, Copy, Clone)]
 pub(crate) struct ClockEventData {
-    pub(crate) delay: u128,
+    pub(crate) delay: u128, // delay is expressed in nanoseconds
     pub(crate) userdata: host::__wasi_userdata_t,
 }
 
--- a/src/hostcalls_impl/misc.rs
+++ b/src/hostcalls_impl/misc.rs
@@ -214,49 +214,74 @@ pub(crate) fn poll_oneoff(
     enc_pointee(memory, nevents, 0)?;
 
     let input_slice = dec_slice_of::<wasm32::__wasi_subscription_t>(memory, input, nsubscriptions)?;
+    let subscriptions = input_slice
+        .into_iter()
+        .map(dec_subscription)
+        .collect::<Result<Vec<_>>>()?;
+    let output_slice = dec_slice_of_mut::<wasm32::__wasi_event_t>(memory, output, nsubscriptions)?;
+    let mut output_slice_iter = output_slice.iter_mut();
+    let mut events_count = 0;
 
     let mut timeout: Option<ClockEventData> = None;
     let mut fd_events = Vec::new();
-    for event in input_slice.iter().map(dec_subscription) {
-        let event = event?;
-        match event.type_ {
+    for subscription in subscriptions {
+        match subscription.type_ {
             host::__WASI_EVENTTYPE_CLOCK => {
-                let clock = unsafe { event.u.clock };
+                let clock = unsafe { subscription.u.clock };
                 let delay = wasi_clock_to_relative_ns_delay(clock)?;
 
                 log::debug!("poll_oneoff event.u.clock = {:?}", clock);
                 log::debug!("poll_oneoff delay = {:?}ns", delay);
 
-                let current_event = ClockEventData {
+                let current = ClockEventData {
                     delay,
-                    userdata: event.userdata,
+                    userdata: subscription.userdata,
                 };
-                let timeout_event = timeout.get_or_insert(current_event);
-                if current_event.delay < timeout_event.delay {
-                    *timeout_event = current_event;
+                let timeout = timeout.get_or_insert(current);
+                if current.delay < timeout.delay {
+                    *timeout = current;
                 }
             }
             type_
                 if type_ == host::__WASI_EVENTTYPE_FD_READ
                     || type_ == host::__WASI_EVENTTYPE_FD_WRITE =>
             {
-                let wasi_fd = unsafe { event.u.fd_readwrite.fd };
+                let wasi_fd = unsafe { subscription.u.fd_readwrite.fd };
                 let rights = if type_ == host::__WASI_EVENTTYPE_FD_READ {
                     host::__WASI_RIGHT_FD_READ
                 } else {
                     host::__WASI_RIGHT_FD_WRITE
                 };
-                let descriptor = unsafe {
-                    &wasi_ctx
-                        .get_fd_entry(wasi_fd, rights, 0)?
-                        .fd_object
-                        .descriptor
-                };
-                fd_events.push(FdEventData {
-                    descriptor,
-                    type_: event.type_,
-                    userdata: event.userdata,
-                })
+
+                match unsafe {
+                    wasi_ctx
+                        .get_fd_entry(wasi_fd, rights, 0)
+                        .map(|fe| &fe.fd_object.descriptor)
+                } {
+                    Ok(descriptor) => fd_events.push(FdEventData {
+                        descriptor,
+                        type_: subscription.type_,
+                        userdata: subscription.userdata,
+                    }),
+                    Err(err) => {
+                        let event = host::__wasi_event_t {
+                            userdata: subscription.userdata,
+                            type_,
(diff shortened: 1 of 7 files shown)


kubkon:
OK, this PR is ready for review. ~~Note however that it is (unfortunately) expected to fail as it makes breaking API changes which still need to be addressed in `wasmtime`.~~ The PR is expected to pass the tests now :-)

Additionally, this PR is blocked by CraneStation/wasi-misc-tests#38 (i.e., it really should include `misc_testsuite` fast-forward to the latest upstream when CraneStation/wasi-misc-tests#38 lands).

src/hostcalls_impl/misc.rs
```diff
@@ -205,9 +205,15 @@ pub(crate) fn poll_oneoff(
...
-    let input: Vec<_> = input_slice.iter().map(dec_subscription).collect();
+
+    let events = hostcalls_impl::poll_oneoff(wasi_ctx, input_slice.iter().map(dec_subscription))?;
+
     let output_slice = dec_slice_of_mut::<wasm32::__wasi_event_t>(memory, output, nsubscriptions)?;
-    let events_count = hostcalls_impl::poll_oneoff(input, output_slice)?;
+    let mut events_count = 0;
+    for (i, event) in events.into_iter().enumerate() {
+        output_slice[i] = enc_event(event);
+        events_count += 1;
```
marmistrz: You can just use `events.len()` and avoid mutability. And please avoid the indices, they're not necessary here (the original code also avoids them). See also how it's done in #140.

src/sys/unix/hostcalls_impl/misc.rs
```diff
@@ -70,78 +69,127 @@ pub(crate) fn clock_time_get(clock_id: host::__wasi_clockid_t) -> Result<host::_
...
-        .filter_map(|event| match event {
-            Ok(event) if event.type_ == wasm32::__WASI_EVENTTYPE_CLOCK => Some(ClockEventData {
-                delay: wasi_clock_to_relative_ns_delay(unsafe { event.u.clock }).ok()? / 1_000_000,
-                userdata: event.userdata,
-            }),
-            _ => None,
-        })
-        .min_by_key(|event| event.delay);
+    wasi_ctx: &WasiCtx,
+    input: impl Iterator<Item = Result<host::__wasi_subscription_t>>,
```
marmistrz: Can we accept `Vec<FdEventData>` and  `Option<ClockEventData>` for timeout as in #140? The conversion code cross-platform code and should not reside in this module.

src/sys/unix/hostcalls_impl/misc.rs
```diff
@@ -70,78 +69,127 @@ pub(crate) fn clock_time_get(clock_id: host::__wasi_clockid_t) -> Result<host::_
...
-        .iter()
-        .filter_map(|event| match event {
-            Ok(event)
-                if event.type_ == wasm32::__WASI_EVENTTYPE_FD_READ
-                    || event.type_ == wasm32::__WASI_EVENTTYPE_FD_WRITE =>
-            {
-                Some(FdEventData {
-                    fd: unsafe { event.u.fd_readwrite.fd } as c_int,
+    let mut timeout: Option<ClockEventData> = None;
+    let mut fd_events = Vec::new();
```
marmistrz: See the comment above.

src/sys/unix/hostcalls_impl/misc.rs
```diff
@@ -70,78 +69,127 @@ pub(crate) fn clock_time_get(clock_id: host::__wasi_clockid_t) -> Result<host::_
...
                 // Nothing else has been defined in the specification, and these are also the only two
                 // events we filtered before. If we get something else here, the code has a serious bug.
                 _ => unreachable!(),
             };
-            nix::poll::PollFd::new(event.fd, flags)
+            PollFd::new(event.fd, flags)
         })
         .collect();
+
     let timeout = timeout.map(|ClockEventData { delay, userdata }| ClockEventData {
```
marmistrz: We can remove this map and just use `try_into` as in #140.

src/sys/unix/hostcalls_impl/misc.rs
```diff
@@ -161,116 +209,106 @@ fn wasi_clock_to_relative_ns_delay(
...
                 },
             },
-        };
-        output_slice[0] = enc_event(output_event);
-        1
+        }]
     } else {
         // shouldn't happen
-        0
+        vec![]
```
marmistrz: Then accept `ClockEventData`, not an `Option<ClockEventData>`. Accepting `Option<ClockEventData>` suggests that `None` is a valid value. See how #140 does it.

marmistrz (changes requested):
I closed #140, let's keep all the refactoring in one PR.

kubkon:
> I closed #140, let's keep all the refactoring in one PR.

That's my bad at poorly synchronising efforts between the two of us. Apologies!

src/sys/unix/hostcalls_impl/misc.rs
kubkon: ~~Ah, good idea. Let me refactor it following #140 then!~~ Actually, on second thought, I'd like to keep this PR focused on fixing a bug in `poll_oneoff`. Would you mind handling more refactoring in a subsequent PR?

marmistrz: I'd prefer to have one decent refactor commit, but if this is a major blocker and should be merged ASAP, I'm fine with finishing the refactor in a subsequent PR.

kubkon: I guess we're not in such a hurry, so let's see how we get on with it :-)

src/sys/unix/hostcalls_impl/misc.rs
```diff
@@ -70,78 +69,127 @@ pub(crate) fn clock_time_get(clock_id: host::__wasi_clockid_t) -> Result<host::_
...
+                let timeout_event = timeout.get_or_insert(current_event);
+                if current_event.delay < timeout_event.delay {
+                    *timeout_event = current_event;
+                }
+            }
+            host::__WASI_EVENTTYPE_FD_READ => {
+                let wasi_fd = unsafe { event.u.fd_readwrite.fd };
+                let fd = unsafe {
+                    wasi_ctx
+                        .get_fd_entry(wasi_fd, host::__WASI_RIGHT_FD_READ, 0)?
```
dunnock: This line seems throwing error when at least one bad wasi file descriptor supplied. Should not it rather return event for relevant record with error code EBADF? Since there might be mix of polls subscriptions and some might be valid, while some invalid it is important to return those which are valid too.

I've added relevant test https://github.com/kubkon/wasi-misc-tests/pull/2/commits/ed16186b8d4a762020083c8b36d91b12a96f9580

src/hostcalls_impl/misc.rs
kubkon: Good point on `events.len()`, thanks! I'm less clear on "avoiding the indices". Are you saying that calling

```rust
*output_slice_cur.next().unwrap() = enc_event(event);
```

is safer than using indices because AFAIK, in this case, both will panic if we exceed the range. Or is it a stylistic concern?

marmistrz: Both. Probably with an `expect` instead of `unwrap`. It's cleaner and it's not possible to use the wrong variable as the index.

src/sys/unix/hostcalls_impl/misc.rs
kubkon: OK, I've now had a closer look at this. We cannot accept `Vec<FdEventData>` as this will be host specific (on *nix, we deal with file descriptors, on Windows it will be handles). However, if we modify `FdEventData` to contain our `Descriptor` struct, then this should be possible to refactor out.

src/sys/unix/hostcalls_impl/misc.rs
kubkon: Yep, I see it now, thanks!

kubkon: Now that I've included the smoke test CraneStation/wasi-misc-tests#38 in this PR, this seems to be the only (I hope I didn't miss anything!) major thing left before this PR is ready for merging.

kubkon: Here's a question about the spec; would love your comment about this one @sunfishcode. From what I see in the implementation and the spec, the user can link subscriptions with events by the field `userdata` which is present in both and set to the same, user-specified value. Does that mean that we don't have to preserve the order of events relative to the order of the (input) subscriptions? I guess this would make sense especially since in the spec we do say "concurrently poll" [WASI-api#__wasi_poll_oneoff]. And a follow-up question. If indeed `userdata` is used as an identifier for `(subscription, event)` pair, we (purposely) do not enforce uniqueness of `userdata` among the input subscriptions which could very well lead to collisions in the output events, but I guess it is up to the client to enforce the uniqueness?

[WASI-api#__wasi_poll_oneoff]: https://github.com/CraneStation/wasmtime/blob/master/docs/WASI-api.md#__wasi_poll_oneoff

marmistrz:
@kubkon can we have a more descriptive PR title?

src/sys/unix/hostcalls_impl/misc.rs
kubkon: OK, I think I now have convinced myself that the order in general will not be preserved nor is expected to be.

kubkon:
I think we're in a pretty good shape with this now, so I'm gonna go ahead and merge this. If you guys feel that we should still modify something, or something isn't clear, please feel free to submit an issue, and I'll be happy to fix anything that is remaining :-)

PR #138, opened 2019-10-22
Sync with latest wasmtime rev

kubkon:
This commit syncs tests with latest wasmtime revision.
As such, it now utilises the `wasmtime-api` crate for
runtime setup.

Closes #126, #127, #128, #129.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,14 +32,12 @@ winx = { path = "winx" }
 winapi = "0.3"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


sunfishcode (approved):
Thanks for moving this over to wasmtime-api while here; that's the direction wasmtime is generally heading!

kubkon:
> Thanks for moving this over to wasmtime-api while here; that's the direction wasmtime is generally heading!

Yep, it was long overdue! :-)

PR #139, opened 2019-10-22
[WIP] first updates to introduce sockets

ibaryshnikov:
It's an attempt to implement two first networking methods:
- `sock_socket`, which creates a file descriptor and stores some information like domain, type, protocol
- `sock_connect`, which opens connection using fd obtained in `sock_socket`

### DIFF
--- a/src/fdentry.rs
+++ b/src/fdentry.rs
@@ -2,11 +2,20 @@ use crate::sys::fdentry_impl::{determine_type_and_access_rights, OsFile};
 use crate::{host, Error, Result};
 use std::mem::ManuallyDrop;
 use std::path::PathBuf;
-use std::{fs, io};
+use std::{fs, io, net};
+
+#[derive(Debug)]
+pub struct SocketDetails {
+    pub socket_domain: i32,
+    pub socket_type: u8,
+    pub socket_protocol: i32,
+}
 
 #[derive(Debug)]
 pub(crate) enum Descriptor {
     OsFile(OsFile),
+    Socket(net::TcpStream),
+    SocketFd(SocketDetails), // stores allocated fd, and options: domain, type and protocol
     Stdin,
     Stdout,
     Stderr,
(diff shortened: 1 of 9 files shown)


src/fdentry.rs
```diff
@@ -2,11 +2,20 @@ use crate::sys::fdentry_impl::{determine_type_and_access_rights, OsFile};
 use crate::{host, Error, Result};
 use std::mem::ManuallyDrop;
 use std::path::PathBuf;
-use std::{fs, io};
+use std::{fs, io, net};
+
+#[derive(Debug)]
+pub struct SocketDetails {
+    pub socket_domain: i32,
+    pub socket_type: u8,
```
steveej, line 10: @ibaryshnikov have you considered an enum type here instead? What would be the down/upsides to that?

marmistrz:
If you're introducing socket support, you probably should also take care of this issue: https://github.com/CraneStation/wasi-common/issues/167

kubkon:
Hey @ibaryshnikov! Would it be OK if you continued your work directly in `wasmtime` repo after CraneStation/wasmtime#502 lands?

ibaryshnikov:
@kubkon yes, totally fine. I'll need to open a new PR directly in wasmtime, right?

src/fdentry.rs
ibaryshnikov: I guess it may be useful, I'd like to try. It's a common pattern while working with FFI to provide safe wrappers

ibaryshnikov:
@marmistrz thanks, I'll take a look

kubkon:
> @kubkon yes, totally fine. I'll need to open a new PR directly in wasmtime, right?

Yep, that's the idea :-)

kubkon:
I'm gonna go ahead and close this PR. @ibaryshnikov please reopen it directly in [wasmtime](https://github.com/cranestation/wasmtime).

PR #146, opened 2019-10-25
make `host` module public

pchickey:
In `lucet` we use some of the constants defined in (the predecessor in `lucet-wasi` of) this module. This PR makes `wasi_common::host` public and converts all of the definitions in that module from `pub(crate)` to `pub`.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -6,195 +6,195 @@
 use crate::{Error, Result};
 use std::{io, slice, str};
 
-pub(crate) type void = ::std::os::raw::c_void;
+pub type void = ::std::os::raw::c_void;
 
-pub(crate) type __wasi_advice_t = u8;
-pub(crate) const __WASI_ADVICE_NORMAL: __wasi_advice_t = 0;
-pub(crate) const __WASI_ADVICE_SEQUENTIAL: __wasi_advice_t = 1;
-pub(crate) const __WASI_ADVICE_RANDOM: __wasi_advice_t = 2;
-pub(crate) const __WASI_ADVICE_WILLNEED: __wasi_advice_t = 3;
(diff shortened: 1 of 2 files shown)


sunfishcode:
I have some ideas for how we could refactor `host`, but I expect the basic functionality it provides will continue to be available in any case, so this seems fine to me.

kubkon:
Hmm, would it be possible to use `wasm32` instead of `host`? I specifically made `host` private (it used to be public) so that we can put a clear line between functionality that is internal to `wasi-common` and that which is not. Also, this would simplify merging of `wasm32` and `host` into one. If you definitely need `host` to be public, then perhaps merging of the two is not such a trivial as I initially assumed.

marmistrz:
IMO the distinction between `wasm32` and `host` is quite confusing. Probably it's a good idea to mention some rules of a thumb when which of those modules should be used.

kubkon:
@marmistrz To provide you with some context, I've had this chat with @sunfishcode a while back, and we generally agreed that there is little reason to have two separate modules `host` and `wasm32` which essentially expose the same constants, structs, etc. Hence, there is a plan to unify the two, we just didn't get around to it yet ;-)

sunfishcode:
https://github.com/CraneStation/wasi-common/pull/151 is now a PR showing one possible approach.

pchickey:
I'm going to close this - I got rid of lucet's dependency on this module temporarily. I can add it back in once we've engineered a way forward.

Right now I really like the way #152 is moving this forward - we really should be able to generate both the C ABI level constants / type aliases, and idiomatic rust structs / enums, and the translation routines between them, all from the witx descriptions of WASI. That will be a big improvement for this library and validate that the effort we're putting into witx is worthwhile :)

PR #147, opened 2019-10-25
WasiCtxBuilder: add methods to provide Files each for stdio

pchickey:
In `lucet-wasi` and `lucet-wasi-fuzz` we have tests that redirect stdio to pipes, so we can programmatically feed input or output to a WasiCtx. This PR exposes methods on `WasiCtxBuilder` allowing arbitrary `File`s to be used for each stdio stream.


sunfishcode (approved):
Yes, that makes sense.

src/ctx.rs
```diff
@@ -96,6 +96,24 @@ impl WasiCtxBuilder {
         Ok(self)
     }
 
+    /// Provide a File to use as stdin
+    pub fn stdin(mut self, file: File) -> Result<Self> {
+        self.fds.insert(0, FdEntry::from(file)?);
```
kubkon, line 101: Hmm, imagine a situation where we've already called `pub fn stdin(...)` on `WasiCtx` and inserted some `File` as our `stdin` handle. By default, upon insertion, we'll mark the `FdObject.needs_close == true`. But then suppose we somehow somewhere change our mind about it, and mark it as `FdObject.needs_close = false`. In this case, if we ever call `pub fn stdin(...)` again, we will remove that `FdObject` from `self.fds` but when it goes out of scope, we will not drop it automatically as it was marked as `needs_close == false` and it's behind a `ManuallyDrop` wrapper. The question here is should we worry about this? Or would it be intended behaviour?

kubkon: BTW, I'm just thinking out loud here, so if you feel that my comment is nonsense, please do flag it up!

sunfishcode: We don't currently have any code which sets `needs_close` to false -- do we still need the `needs_close` field? But even if we do, any code which sets it to false would need to ensure that the descriptor's ownership has been transferred somewhere else.

kubkon: Good point, I guess we can get rid of altogether.

kubkon (approved):

pchickey:
Thanks! I can switch lucet over to using wasi-common now, I just need a new version released to crates.io that includes this change. Would you mind cutting a release sometime soon?

PR #149, opened 2019-10-25
Remove `needs_close` from `FdObject`

kubkon:
This commit removes `needs_close` field from `FdObject`, and stores the underlying `Descriptor` without the `ManuallyDrop` wrapper.

As per our discussion in #147.

### DIFF
--- a/src/fdentry.rs
+++ b/src/fdentry.rs
@@ -1,6 +1,5 @@
 use crate::sys::fdentry_impl::{determine_type_and_access_rights, OsFile};
 use crate::{host, Error, Result};
-use std::mem::ManuallyDrop;
 use std::path::PathBuf;
 use std::{fs, io};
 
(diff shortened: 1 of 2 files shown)


sunfishcode (approved):
I'm trying to decide if we'll ever want the ability to have `WasiCtx` hold on to file descriptors it doesn't own, and if so, if it's worth keeping this code around until then.

I think where I'm arriving at is, we may want this capability at some point, but I think we're going to want it in a different form. Instead of making the `Descriptor` be `ManuallyDrop`, it should be the `OsFile` or perhaps even the underlying `std::fs::File`. It's ok to drop a `Stdin` object, for example. So I think removing it for now makes sense; we can reintroduce it in a different way in the future if we need it.

kubkon:
> I'm trying to decide if we'll ever want the ability to have `WasiCtx` hold on to file descriptors it doesn't own, and if so, if it's worth keeping this code around until then.
> 
> I think where I'm arriving at is, we may want this capability at some point, but I think we're going to want it in a different form. Instead of making the `Descriptor` be `ManuallyDrop`, it should be the `OsFile` or perhaps even the underlying `std::fs::File`. It's ok to drop a `Stdin` object, for example. So I think removing it for now makes sense; we can reintroduce it in a different way in the future if we need it.

I agree that we can always re-add it in the future if need be, and make it slightly more informed given the ever-improving state of the crate.

PR #151, opened 2019-10-26
Reorganize host.rs and wasm32.rs.

sunfishcode:
Reorganize host.rs and wasm32.rs into host.rs, wasi.rs, and wasi32.rs.
    
Most of the contents of host.rs was not actually host-specific, as most
of the types are fixed-size types like u32 or i64. These types are now
in wasi.rs.
    
The few types which do have pointer or usize-sized values now remain,
in two versions: host.rs has versions which use actual raw pointers and
usize, and wasi32.rs has versions which use u32 to represent them.

I'm not sure this is the right approach, and welcome comments here!

This approach does not provide a type-level distinction between most
host types and wasm types, though this isn't new; the previous code
also didn't make a strong distinction, because eg. `host::__wasi_filetype_t`
was the same type as `wasm32::__wasi_filetype_t`.


kubkon (approved):
This looks great, thanks @sunfishcode! I especially like the fact that we're already factoring in the fact that in the future we expect 64bit WASI.

marmistrz (approved):
Sounds great! I really like the fact that those unclear cases need to be explicitly opted-in by using either host or wasi32.

One nitpick: how about removing `use crate::wasi::*` from `wasi32.rs` and `host.rs` so that `wasi.rs` and X are disjoint and that there's only one way of using the cross-platform types?

It would probably also be a good idea to add some comment about what the difference between `host` and `wasi32` actually is.

kubkon:
> Sounds great! I really like the fact that those unclear cases need to be explicitly opted-in by using either host or wasi32.
> 
> One nitpick: how about removing `use crate::wasi::*` from `wasi32.rs` and `host.rs` so that `wasi.rs` and X are disjoint and that there's only one way of using the cross-platform types?
> 
> It would probably also be a good idea to add some comment about what the difference between `host` and `wasi32` actually is.

Hmm, I think I actually prefer the way it is now, since, as far as I understand, for the lib user it is now a matter of importing either `wasi_common::wasi32` or `wasi_common::wasi64` as both (would) include the common `wasi` module, rather than importing `wasi_common::wasi32` and `wasi_common::wasi` separately. A matter of taste I guess ;-)

When it comes to having `crate::wasi::*` inside of the `host` module, I'm fine with excluding the import here as the `host` module should only be used for internal purposes (I think).

kubkon:
> > Sounds great! I really like the fact that those unclear cases need to be explicitly opted-in by using either host or wasi32.
> > One nitpick: how about removing `use crate::wasi::*` from `wasi32.rs` and `host.rs` so that `wasi.rs` and X are disjoint and that there's only one way of using the cross-platform types?
> > It would probably also be a good idea to add some comment about what the difference between `host` and `wasi32` actually is.
> 
> Hmm, I think I actually prefer the way it is now, since, as far as I understand, for the lib user it is now a matter of importing either `wasi_common::wasi32` or `wasi_common::wasi64` as both (would) include the common `wasi` module, rather than importing `wasi_common::wasi32` and `wasi_common::wasi` separately. A matter of taste I guess ;-)
> 
> When it comes to having `crate::wasi::*` inside of the `host` module, I'm fine with excluding the import here as the `host` module should only be used for internal purposes (I think).

Actually, I misread this. Unless we make `use crate::wasi::*` public, I agree with @marmistrz that it might make more sense to make `wasi32` and `host` disjoint by removing the import altogether.

src/wasi32.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+pub(crate) const INTPTR_MIN: i32 = -2147483648;
+pub(crate) const INTPTR_MAX: u32 = 2147483647;
+pub(crate) const UINTPTR_MAX: u32 = 4294967295;
+pub(crate) const PTRDIFF_MIN: i32 = -2147483648;
+pub(crate) const PTRDIFF_MAX: u32 = 2147483647;
+pub(crate) const SIG_ATOMIC_MIN: i32 = -2147483648;
+pub(crate) const SIG_ATOMIC_MAX: u32 = 2147483647;
+pub(crate) const SIZE_MAX: u32 = 4294967295;
```
kubkon: Yet another thing that slipped through the radar, shouldn't these be public? I'm not sure whether all of them, but type aliases such as `uintptr_t` for sure as it's needed in order to plug in the lib into the runtime, for instance, as is the case in [wasmtime_wasi#L166].

[wasmtime_wasi#L166]: https://github.com/CraneStation/wasmtime/blob/master/wasmtime-wasi/src/syscalls.rs#L166

kubkon (changes requested):

jedisct1:
The `wasm` and `host` versions of fixed-size types had the advantage of making a distinction between what has to be encoded as little-endian vs what may use a different encoding.

sunfishcode:
> One nitpick: how about removing use crate::wasi::* from wasi32.rs and host.rs so that wasi.rs and X are disjoint and that there's only one way of using the cross-platform types?

It a plain `use`, not a `pub use`, so iirc it's not exported. host.rs and wasi32.rs do depend on things declared in wasi.rs, so they need the import, but they don't export it to their users. There's only one way to use the cross-platforms types, and that's `wasi::`.

> Yet another thing that slipped through the radar, shouldn't these be public? I'm not sure whether all of them, but type aliases such as uintptr_t for sure as it's needed in order to plug in the lib into the runtime, for instance, as is the case in wasmtime_wasi#L166.

> The wasm and host versions of fixed-size types had the advantage of making a distinction between what has to be encoded as little-endian vs what may use a different encoding.

There are a few different ways to could approach this, but I'd like to propose the following approach:

Types like `__wasi_fd_t` are always in host-endian order. The only place we should have endianness conversions happening is in src/memory.rs when we're loading out of a wasm linear memory (which is explicitly little-endian), and while this isn't the case yet, it shouldn't cast into `__wasi_fd_t` and other types until it's done the conversion. Hopefully soon, most of src/memory.rs will be auto-generated, which should make this feasible.

Does that make sense?

sunfishcode:
Oops, forgot to actually reply to this:
> Yet another thing that slipped through the radar, shouldn't these be public? I'm not sure whether all of them, but type aliases such as uintptr_t for sure as it's needed in order to plug in the lib into the runtime, for instance, as is the case in wasmtime_wasi#L166.

I believe this is sorted out in the rebase I just pushed.

sunfishcode:
With the latest push, I've now added a patch implementing [this idea](https://github.com/CraneStation/wasi-common/pull/151#issuecomment-547651359), of fully encapsulating all endianness issues in memory.rs.

I haven't tested it on a big-endian platform, so it's possible there are endianness bugs (though the code before this patch definitely has endianness bugs), however now we have a principled and simple approach, and we can easily fix any bugs that are found.

src/host.rs
```diff
@@ -1,1049 +1,73 @@
...
+pub(crate) type void = ::std::os::raw::c_void;
 
 #[repr(C)]
 #[derive(Copy, Clone)]
-pub(crate) struct __wasi_prestat_t {
-    pub(crate) pr_type: __wasi_preopentype_t,
-    pub(crate) u: __wasi_prestat_t___wasi_prestat_u,
+pub struct __wasi_prestat_t {
+    pub pr_type: __wasi_preopentype_t,
+    pub u: __wasi_prestat_t___wasi_prestat_u,
```
kubkon: Is there a reason why these are `pub` and not `pub(crate)`? AFAICS, `mod host` is private in `lib.rs`.

src/host.rs
```diff
@@ -1,1049 +1,73 @@
...
 }
 
 #[repr(C)]
 #[derive(Copy, Clone)]
-pub(crate) union __wasi_prestat_t___wasi_prestat_u {
-    pub(crate) dir: __wasi_prestat_t___wasi_prestat_u___wasi_prestat_u_dir_t,
+pub union __wasi_prestat_t___wasi_prestat_u {
+    pub dir: __wasi_prestat_t___wasi_prestat_u___wasi_prestat_u_dir_t,
```
kubkon: Same here.

src/host.rs
```diff
@@ -1,1049 +1,73 @@
...
     _bindgen_union_align: u64,
 }
 
 #[repr(C)]
 #[derive(Debug, Copy, Clone)]
-pub(crate) struct __wasi_prestat_t___wasi_prestat_u___wasi_prestat_u_dir_t {
-    pub(crate) pr_name_len: usize,
+pub struct __wasi_prestat_t___wasi_prestat_u___wasi_prestat_u_dir_t {
+    pub pr_name_len: usize,
```
kubkon: Same here.

src/host.rs
```diff
@@ -1,1049 +1,73 @@
...
 #[repr(C)]
 #[derive(Debug, Copy, Clone)]
-pub(crate) struct __wasi_fdstat_t {
-    pub(crate) fs_filetype: __wasi_filetype_t,
-    pub(crate) fs_flags: __wasi_fdflags_t,
-    pub(crate) fs_rights_base: __wasi_rights_t,
-    pub(crate) fs_rights_inheriting: __wasi_rights_t,
+pub struct __wasi_ciovec_t {
+    pub buf: *const void,
+    pub buf_len: usize,
```
kubkon: Same here.

src/host.rs
```diff
@@ -1,1049 +1,73 @@
...
-    pub(crate) flags: __wasi_subclockflags_t,
-}
-
-#[repr(C)]
-#[derive(Debug, Copy, Clone)]
-pub(crate) struct __wasi_subscription_t___wasi_subscription_u___wasi_subscription_u_fd_readwrite_t {
-    pub(crate) fd: __wasi_fd_t,
+pub struct __wasi_iovec_t {
+    pub buf: *mut void,
+    pub buf_len: usize,
```
kubkon: Same here.

src/host.rs
```diff
@@ -1,1049 +1,73 @@
...
 }
 
 #[allow(unused)]
-pub(crate) unsafe fn ciovec_to_host(ciovec: &__wasi_ciovec_t) -> io::IoSlice {
+pub unsafe fn ciovec_to_host(ciovec: &__wasi_ciovec_t) -> io::IoSlice {
```
kubkon: Same here.

src/host.rs
```diff
@@ -1,1049 +1,73 @@
...
     let slice = slice::from_raw_parts(ciovec.buf as *const u8, ciovec.buf_len);
     io::IoSlice::new(slice)
 }
 
 #[allow(unused)]
-pub(crate) unsafe fn ciovec_to_host_mut(ciovec: &mut __wasi_ciovec_t) -> io::IoSliceMut {
+pub unsafe fn ciovec_to_host_mut(ciovec: &mut __wasi_ciovec_t) -> io::IoSliceMut {
```
kubkon: Same here.

src/host.rs
```diff
@@ -1,1049 +1,73 @@
...
     let slice = slice::from_raw_parts_mut(ciovec.buf as *mut u8, ciovec.buf_len);
     io::IoSliceMut::new(slice)
 }
 
-pub(crate) unsafe fn iovec_to_host(iovec: &__wasi_iovec_t) -> io::IoSlice {
+pub unsafe fn iovec_to_host(iovec: &__wasi_iovec_t) -> io::IoSlice {
```
kubkon: Same here.

src/host.rs
```diff
@@ -1,1049 +1,73 @@
...
     let slice = slice::from_raw_parts(iovec.buf as *const u8, iovec.buf_len);
     io::IoSlice::new(slice)
 }
 
-pub(crate) unsafe fn iovec_to_host_mut(iovec: &mut __wasi_iovec_t) -> io::IoSliceMut {
+pub unsafe fn iovec_to_host_mut(iovec: &mut __wasi_iovec_t) -> io::IoSliceMut {
```
kubkon: Same here.

src/wasi.rs
```diff
@@ -1,9 +1,6 @@
...
-//! still largely reflects that, however it's been heavily modified, to
-//! be host-independent, to avoid exposing libc implementation details,
-//! to clean up cases where the headers use complex preprocessor macros,
-//! and to
+//! Types and constants shared between 32-bit and 64-bit wasi. Types involving
+//! pointer or `usize`-sized data are excluded here, so this file only contains
+//! fixed-size types, so it's host/target independent.
 
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
```
kubkon, line 6: I believe `#![allow(dead_code)]` is no longer needed in the line below. Could you verify and remove it while we're here?

src/wasi32.rs
```diff
@@ -0,0 +1,232 @@
+//! Types and constants specific to 32-bit wasi. These are similar to the types
+//! in the `host` module, but pointers and `usize` values are replaced with
+//! `u32`-sized types.
+
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+#![allow(dead_code)]
```
kubkon, line 7: I believe `#![allow(dead_code)]` is no longer needed. Could you verify and remove it while we're here?

kubkon (approved):
LGTM, just a couple of minor nitpicks!

src/host.rs
sunfishcode: Indeed, I've now made these all be `pub(crate)`.

kubkon:
And of course I forgot to check the Windows platform for the unused code. Perhaps in those 4 instances that caused the CI to fail on Windows:

```shell
error: constant item is never used: `RIGHTS_BLOCK_DEVICE_BASE`
   --> src\wasi.rs:152:1
    |
152 | pub(crate) const RIGHTS_BLOCK_DEVICE_BASE: __wasi_rights_t = RIGHTS_ALL;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D dead-code` implied by `-D warnings`

error: constant item is never used: `RIGHTS_BLOCK_DEVICE_INHERITING`
   --> src\wasi.rs:153:1
    |
153 | pub(crate) const RIGHTS_BLOCK_DEVICE_INHERITING: __wasi_rights_t = RIGHTS_ALL;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: constant item is never used: `RIGHTS_CHARACTER_DEVICE_BASE`
   --> src\wasi.rs:154:1
    |
154 | pub(crate) const RIGHTS_CHARACTER_DEVICE_BASE: __wasi_rights_t = RIGHTS_ALL;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: constant item is never used: `RIGHTS_CHARACTER_DEVICE_INHERITING`
   --> src\wasi.rs:155:1
    |
155 | pub(crate) const RIGHTS_CHARACTER_DEVICE_INHERITING: __wasi_rights_t = RIGHTS_ALL;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

we could add `#[allow(unused)]` next to each item? What are your thoughts on this? Ultimately, when we fill in the blanks in the Windows impl, all of the above constants will be used :-)

src/wasi.rs
sunfishcode: Removing it broke the Windows build, so I've reinstated it for now.

sunfishcode:
Oops, I didn't see your comment here. I'll clean that up in a followup PR.

kubkon:
> Oops, I didn't see your comment here. I'll clean that up in a followup PR.

Np! :-)

PR #152, opened 2019-10-26
Use generated type bindings

sunfishcode:
This is based on #151 and is somewhat experimental. It also depends on at least https://github.com/WebAssembly/WASI/pull/129.

Use the witx API descriptions to generate the bulk of the contents of
host.rs, wasi.rs, and wasi32.rs.
    
This also prunes out many of the miscellaneous libc definitions from
those files which aren't currently in use by wasi-common. If there's
anything removed that's still needed by someone, it's easy to add things
back in.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v1
+      with:
+        submodules: recursive
     - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
@@ -52,6 +54,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v1
+      with:
+        submodules: recursive
     - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
@@ -99,6 +103,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v1
+      with:
+        submodules: recursive
     - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "misc_testsuite"]
 	path = misc_testsuite
 	url = https://github.com/cranestation/wasi-misc-tests
+[submodule "WASI"]
+	path = WASI
+	url = https://github.com/WebAssembly/WASI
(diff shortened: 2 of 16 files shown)


kubkon:
After briefly investigating, I've tracked the tests failure down to incorrect encoding of `__WASI_EVENTTYPE_CLOCK`, `__WASI_EVENTTYPE_FD_READ` and `__WASI_EVENTTYPE_FD_WRITE` in [CraneStation/rust-wasi] crate compared to `witx`.

In [CraneStation/rust-wasi]:

```rust
pub const __WASI_EVENTTYPE_CLOCK: __wasi_eventtype_t = 0;
pub const __WASI_EVENTTYPE_FD_READ: __wasi_eventtype_t = 1;
pub const __WASI_EVENTTYPE_FD_WRITE: __wasi_eventtype_t = 2;
```

In `witx`:

```rust
pub const __WASI_EVENTTYPE_CLOCK: __wasi_eventtype_t = 1;
pub const __WASI_EVENTTYPE_FD_READ: __wasi_eventtype_t = 2;
pub const __WASI_EVENTTYPE_FD_WRITE: __wasi_eventtype_t = 4;
```

It seems that `witx` encodes them as an 8-bit flag hence the assumed values of `1, 2, 4`. It seems that the `witx` route was always the intended route after quickly analysing the old API in [WebAssembly/WASI `witx` spec], so I've gone ahead and submitted a patch to [CraneStation/rust-wasi]. This can be tracked here: CraneStation/rust-wasi#12.

[CraneStation/rust-wasi]: https://github.com/CraneStation/rust-wasi
[WebAssembly/WASI `witx` spec]: https://github.com/WebAssembly/WASI/blob/master/phases/old/witx/typenames.witx#L483

sunfishcode:
Oops, we have a slight race condition here :-}. This turned out to be a bug in the witx files; the event type values shouldn't be flags. I've now submitted a WASI repo PR to fix this.

kubkon:
> Oops, we have a slight race condition here :-}. This turned out to be a bug in the witx files; the event type values shouldn't be flags. I've now submitted a WASI repo PR to fix this.

Sounds good! For anyone interested in tracking this, here's the link to the WASI PR WebAssembly/WASI#133.

kubkon (approved):
LGTM! When WebAssembly/WASI#133 lands, I think we're good to go with this one :-)

PR #154, opened 2019-10-29
Implement a helper for converting &str to CString.

### LINKED ISSUE #104: Refactor conversion to CString
As reported by @sunfishcode in #103:

> As an aside, this sequence of converting to a CString and mapping errors to EILSEQ happens a lot, so it would make for a good utility function:

```rust
let old_path_cstr = CString::new(resolved_old.path().as_bytes()).map_err(|_| Error::EILSEQ)?;
let new_path_cstr = CString::new(resolved_new.path().as_bytes()).map_err(|_| Error::EILSEQ)?;
```

peterhuene:
This commit implements a simple helper for converting `&str` to `CString` and
mapping to the appropriate WASI error.

It also adds a `to_cstring` helper method in `PathGet` where the conversion was
used the most.

Fixes #104.

### DIFF
--- a/src/hostcalls_impl/fs_helpers.rs
+++ b/src/hostcalls_impl/fs_helpers.rs
@@ -1,7 +1,9 @@
 #![allow(non_camel_case_types)]
+use crate::helpers::str_to_cstring;
 use crate::sys::host_impl;
 use crate::sys::hostcalls_impl::fs_helpers::*;
 use crate::{host, Error, Result};
+use std::ffi::CString;
 use std::fs::File;
 use std::path::{Component, Path};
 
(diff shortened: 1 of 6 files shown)


src/hostcalls_impl/fs_helpers.rs
```diff
@@ -19,6 +21,10 @@ impl PathGet {
     pub(crate) fn path(&self) -> &str {
         &self.path
     }
+
+    pub(crate) fn to_cstring(&self) -> Result<CString> {
```
sunfishcode: My one bikeshed is that `to_cstring` doesn't print the whole `PathGet`, but just the `path` part. So it'd be a little clearer to call this `path_as_cstring` or `path_cstring` or so.

sunfishcode (approved):
Looks good!

src/hostcalls_impl/fs_helpers.rs
marmistrz: Rather `path_to_cstring` instead of `path_as_cstring`. `as` conversions are supposed to be cheap.

`path_cstring` sounds fine.

peterhuene: `path_cstring` works for me.

PR #155, opened 2019-10-29
Make `path_get` return `ENOTDIR` when the base isn't a directory.

sunfishcode:
Return `ENOTDIR` instead of `ENOTCAPABLE` when `*at` functions are
given a non-directory as their base. This is in accordance with POSIX:

https://pubs.opengroup.org/onlinepubs/9699919799/functions/open.html

Currently wasi-libc contains some code to detect such cases and rewrite
`ENOTCAPABLE` to `ENOTDIR`, however it's better for WASI implementations
to just do the right thing in the first place.


kubkon (approved):
Looks great!

PR #158, opened 2019-10-30
Revert "Make `path_get` return `ENOTDIR` when the base isn't a directory."

kubkon:
**DO NOT MERGE YET UNTIL CERTAIN DECISIONS ARE MADE**

Reverts CraneStation/wasi-common#155

It turns out I was too hasty in merging this in. The proposed changes in this #155 unfortunately won't have the desired effect as `ENOTCAPABLE` will be returned before `path_get` gets called, namely, when we query the `WasiCtx` for an appropriate `FdEntry` that's tied to the passed in WASI fd:

```rust
let fe = &wasi_ctx.get_fd_entry(dirfd, rights, 0)?;
```

There, the context checks whether the correct rights are attached to the specified descriptor (in this case, whether `dirfd` is indeed a directory), and if something is amiss, we fail with `ENOTCAPABLE`. The relevant bit of code can be found [here].

[here]: https://github.com/CraneStation/wasi-common/blob/master/src/ctx.rs#L216

### DIFF
--- a/src/hostcalls_impl/fs.rs
+++ b/src/hostcalls_impl/fs.rs
@@ -468,8 +468,10 @@ pub(crate) unsafe fn path_create_directory(
     trace!("     | (path_ptr,path_len)='{}'", path);
 
     let rights = host::__WASI_RIGHT_PATH_OPEN | host::__WASI_RIGHT_PATH_CREATE_DIRECTORY;
-    let fo = &wasi_ctx.get_fd_entry(dirfd, rights, 0)?.fd_object;
-    let resolved = path_get(fo, 0, path, false)?;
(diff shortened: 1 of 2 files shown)


sunfishcode (approved):
Arg. I factored this out of a larger patch that I'm working on, and while I thought I tested it, it appears I didn't actually test it with the new test I wrote. Thanks for catching this.

kubkon:
> Arg. I factored this out of a larger patch that I'm working on, and while I thought I tested it, it appears I didn't actually test it with the new test I wrote. Thanks for catching this.

@sunfishcode No probs. I think we might just about be almost OK. See #159 and let me know what you think :-)

PR #159, opened 2019-10-30
Fix path_get returning ENOTDIR when base not dir

kubkon:
~~This commit makes certain adjustments to `WasiCtx` and `FdEntry` by introducing methods `get_nonvalidated_fd_entry` and `get_nonvalidated_fd_entry_mut` which only check if the `FdEntry` corresponding to the specified raw WASI fd exists in the context object but **without** automatically validating the rights. Thanks to postponing the validation until after the `FdEntry` object is extracted from the context, it is possible to check if the extracted `FdEntry` object is indeed a valid dir descriptor when processing paths in `path_get` helper fn. In essence then, this commit closes #158.~~

~~I've made this PR a draft, so that we can get some discussion going before it's ready for merging into the upstream (if at all). One thing I really don't like about the proposed changes in this PR, is the fact that using `get_nonvalidated_fd_entry` requires the caller to remember to call `FdEntry::validate_rights` manually at a later time instant. We currently do not have a mechanism which would enforce this which can potentially lead to certain nasty security bugs such as operating on an fd which doesn't possess the required rights and yet it's fine since we have used `get_nonvalidated_fd_entry` and forgot to call `FdEntry::validate_rights`. I'd really like to circumvent this potential issue *before* we go ahead and merge this (if at all).~~

~~Also, just so that it's clear, this PR already includes a test case for the behaviour specified in #155. (The source code for the test case can be found here CraneStation/wasi-misc-tests#41.)~~

This PR reshuffles a few things in the internals of `WasiCtx` and `FdEntry`. This serves the purpose of correctly facilitating the proposed behaviour in #155, and, in the process, hopefully, uncluttering some bits of code while advocating safe usage of the introduced abstractions.

The changes:
* remove `FdObject` and flatten its contents into `FdEntry`
* accessing `FdEntry`s from `WasiCtx` context object checks only whether the `FdEntry` exists for the specified key of `__wasi_fd_t` (previously, it also automatically verified that correct rights were passed in)
* validating rights is not performed when dereferencing the `FdEntry` as the `Descriptor`; this essentially postpones the validation allowing for changes in #155 to have the desired effect, while also ensuring that the usage of the abstractions is safe
* adds the test case for behaviour proposed in #155 
* closes #158

### DIFF
--- a/src/fdentry.rs
+++ b/src/fdentry.rs
@@ -58,6 +58,14 @@ impl Descriptor {
     }
 }
 
+/// An abstraction struct serving as a wrapper for a host `Descriptor` object which requires
+/// certain base rights `rights_base` and inheriting rights `rights_inheriting` in order to be
+/// accessed correctly.
+///
+/// Here, the `descriptor` field stores the host `Descriptor` object (such as a file descriptor, or
+/// stdin handle), and accessing it can only be done via the provided `FdEntry::as_descriptor` and
+/// `FdEntry::as_descriptor_mut` methods which require a set of base and inheriting rights to be
+/// specified, verifying whether the stored `Descriptor` object is valid for the rights specified.
 #[derive(Debug)]
 pub(crate) struct FdEntry {
     pub(crate) file_type: host::__wasi_filetype_t,
@@ -121,6 +129,13 @@ impl FdEntry {
         )
     }
 
+    /// Convert this `FdEntry` into a host `Descriptor` object provided the specified
+    /// `rights_base` and `rights_inheriting` rights are set on this `FdEntry` object.
+    ///
+    /// The `FdEntry` can only be converted into a valid `Descriptor` object if
+    /// the specified set of base rights `rights_base`, and inheriting rights `rights_inheriting`
+    /// is a subset of rights attached to this `FdEntry`. The check is performed using
+    /// `FdEntry::validate_rights` method. If the check fails, `Error::ENOTCAPABLE` is returned.
     pub(crate) fn as_descriptor(
         &self,
         rights_base: host::__wasi_rights_t,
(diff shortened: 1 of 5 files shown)


src/fdentry.rs
```diff
@@ -123,14 +112,43 @@ impl FdEntry {
...
+                file_type,
+                descriptor: Descriptor::Stderr,
                 rights_base,
                 rights_inheriting,
                 preopen_path: None,
             },
         )
     }
+
+    pub(crate) fn as_descriptor(
```
sunfishcode, line 139: Could you add documentation comments here (and on the `_mut` version below) explaining that the `rights_*` arguments are the rights you need, and that this returns `Err` if the file descriptor doesn't have those rights?

src/fdentry.rs
```diff
@@ -58,29 +58,22 @@ impl Descriptor {
...
-    pub(crate) file_type: host::__wasi_filetype_t,
-    pub(crate) descriptor: Descriptor,
-    // TODO: directories
-}
-
 #[derive(Debug)]
 pub(crate) struct FdEntry {
-    pub(crate) fd_object: FdObject,
+    pub(crate) file_type: host::__wasi_filetype_t,
+    descriptor: Descriptor,
```
sunfishcode, line 72: Could you add a comment here mentioning that `descriptor` holds the state that requires certain rights to access, and that `as_descriptor` is the way to access this field?

sunfishcode (approved):
Cool, thanks for fixing this! I like the symmetry between `FdEntry`'s rights, file type, and preopen fields being public, with `fd_fdstat_get`, `fd_prestat_get`, etc. requiring no rights.

src/fdentry.rs
kubkon: Done...

src/fdentry.rs
kubkon: ...and done :-)

pchickey:
This looks like the right approach to me :)

kubkon:
@sunfishcode Let me know if you reckon we're a go merging this PR into upstream :-)

sunfishcode:
We're a go :-).

PR #160, opened 2019-10-30
Fix unused function warnings on Windows.

peterhuene:
Unfortunately the helpers added in #154 were only used from non-Windows
implementations, which caused compiler warnings on Windows.

This PR moves the helper to be unix-specific and removes the tiny wrapper
around calling `str_to_cstring` on `PathGet.path`.

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -1,6 +1,5 @@
 use crate::{Error, Result};
 use std::convert::TryInto;
-use std::ffi::CString;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 pub(crate) fn systemtime_to_timestamp(st: SystemTime) -> Result<u64> {
@@ -9,8 +8,4 @@ pub(crate) fn systemtime_to_timestamp(st: SystemTime) -> Result<u64> {
         .as_nanos()
         .try_into()
         .map_err(Into::into) // u128 doesn't fit into u64
-}
-
-pub(crate) fn str_to_cstring(s: &str) -> Result<CString> {
-    CString::new(s.as_bytes()).map_err(|_| Error::EILSEQ)
 }
--- a/src/hostcalls_impl/fs_helpers.rs
+++ b/src/hostcalls_impl/fs_helpers.rs
@@ -1,9 +1,7 @@
 #![allow(non_camel_case_types)]
-use crate::helpers::str_to_cstring;
 use crate::sys::host_impl;
 use crate::sys::hostcalls_impl::fs_helpers::*;
 use crate::{fdentry::FdEntry, host, Error, Result};
-use std::ffi::CString;
 use std::fs::File;
 use std::path::{Component, Path};
 
@@ -20,10 +18,6 @@ impl PathGet {
 
     pub(crate) fn path(&self) -> &str {
         &self.path
-    }
-
-    pub(crate) fn path_cstring(&self) -> Result<CString> {
-        str_to_cstring(&self.path)
     }
 }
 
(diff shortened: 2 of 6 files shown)


src/helpers.rs
```diff
@@ -11,6 +10,9 @@ pub(crate) fn systemtime_to_timestamp(st: SystemTime) -> Result<u64> {
         .map_err(Into::into) // u128 doesn't fit into u64
 }
 
-pub(crate) fn str_to_cstring(s: &str) -> Result<CString> {
+#[cfg(unix)]
```
kubkon: Hmm, if these helpers indeed are only Unix specific, could we place them somewhere in `sys/unix` module instead of putting them in host-independent modules with `#[cfg(unix)]` attribute? If possible, I'd really like to keep host-specific functionality in their respective modules rather than have them spread all over.

src/hostcalls_impl/fs_helpers.rs
```diff
@@ -22,7 +20,10 @@ impl PathGet {
         &self.path
     }
 
-    pub(crate) fn path_cstring(&self) -> Result<CString> {
+    #[cfg(unix)]
```
kubkon: Same comment goes here.

kubkon (changes requested):

src/helpers.rs
peterhuene: Agreed, that's a better place for it.  Although the helper isn't specific to Windows, I don't anticipate the need for this helper since we're not calling Windows APIs that would need it afaik.

src/hostcalls_impl/fs_helpers.rs
peterhuene: Maybe we should just remove this method in favor of `str_to_cstring(foo.path())`?  It's not really buying us a lot.

peterhuene: I've removed this function in favor of just using `str_to_cstring`.

kubkon: This sounds like a good idea actually :-)

kubkon (approved):
Looks great, thanks!

src/sys/unix/linux/hostcalls_impl.rs
```diff
@@ -1,7 +1,7 @@
 use super::osfile::OsFile;
-use crate::helpers::str_to_cstring;
 use crate::hostcalls_impl::PathGet;
 use crate::sys::host_impl;
+use crate::sys::unix::str_to_cstring;
 use crate::{host, Error, Result};
 use nix::libc::{self, c_long, c_void};
 use std::convert::TryInto;
```
kubkon, line 7: Since we're already here, would you mind removing the import `use std::ffi::CString`? With your nice helper function, it seems redundant now and generates a warning on Linux :-)

peterhuene: Whoops, missed that one.  Let me remove.

marmistrz:
Could we set up our CI to deny warnings?

peterhuene:
That would be lovely considering how easy it is to introduce warnings for platforms that contributors aren't using locally.

kubkon:
> Could we set up our CI to deny warnings?

Yep, that's an excellent idea. Let me do that then :-)

peterhuene:
Last push was to rebase off of the recent conflicting PR merge.  I believe this is gtg once green (I'll double check for warnings one last time, though).

kubkon:
BTW @peterhuene, you should have received by now an invite to become a collaborator with write access to this repo :-)

PR #161, opened 2019-10-30
Treat all warnings as errors by default

kubkon:
cc @peterhuene

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,13 +33,9 @@ jobs:
       with:
         command: fmt
         args: --all -- --check
-    - name: Build debug
+    - name: Build
       env:
         RUSTFLAGS: "-D warnings"
-      uses: actions-rs/cargo@v1
-      with:
-        command: build
-    - name: Build release
       uses: actions-rs/cargo@v1
       with:
         command: build
@@ -62,11 +58,12 @@ jobs:
         profile: minimal
         toolchain: ${{ matrix.rust }}
         override: true
-    - name: Build debug
+    - name: Build
       continue-on-error: true
       uses: actions-rs/cargo@v1
       with:
         command: build
+        args: --release
 
   test:
     name: Test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,8 @@ jobs:
         command: fmt
         args: --all -- --check
     - name: Build debug
+      env:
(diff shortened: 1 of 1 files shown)


marmistrz:
I think we should rather only treat warnings as errors during CI and allow them for usual builds (e.g. during prototyping)

kubkon:
> I think we should rather only treat warnings as errors during CI and allow them for usual builds (e.g. during prototyping)

That's a fair point. Let me fix it then.

peterhuene:
I also agree that we should probably limit this to CI as warnings are pretty frequent during the development process.

src/fdentry.rs
```diff
@@ -1,5 +1,5 @@
 use crate::sys::fdentry_impl::{determine_type_and_access_rights, OsFile};
-use crate::{host, Error, Result};
+use crate::{host, wasm32, Error, Result};
```
kubkon: This one is just a temp change to see if the CI detects this correctly and fails. It will be removed before merging :-)

sunfishcode:
Is this enabling clippy, and requiring that all PRs be clippy-clean?

I'm somewhat skeptical of that approach. While the warnings builtin into rustc are carefully crafted to be unambiguously worth fixing, a lot of clippy's warnings are quite subjective, and in other projects, I've seen it become a nuisance, especially when its warnings change as clippy evolves.

peterhuene:
It would, yes.  If we don't want that, perhaps (assuming I have the GitHub actions correct):

```yaml
...
    - name: Build debug
      env:
        RUSTFLAGS: "-D warnings"
      uses: actions-rs/cargo@v1
      with:
        command: build
    - name: Build release
      uses: actions-rs/cargo@v1
      env:
        RUSTFLAGS: "-D warnings"
      with:
        command: build
        args: --release
...
```

kubkon:
I guess it’s worth a try, thanks @peterhuene. I reckon that passing the flags to the compiler only in debug build should be more than enough though.

peterhuene (approved):

marmistrz:
Is there any rationale for doing both debug and release build in the CI?

kubkon:
> Is there any rationale for doing both debug and release build in the CI?

Good question. I guess we don't really need the debug build especially since we're running tests later on anyhow (albeit with a couple more dependencies than if invoking `cargo build`). Anyhow, I'm happy to strip away the debug build :-)

PR #163, opened 2019-11-02
Remove `RIGHTS_SHARED_MEMORY_BASE`.

sunfishcode:
This came from CloudABI, but WASI doesn't include POSIX-style
shared-memory objects.


kubkon (approved):
Sounds good!

PR #165, opened 2019-11-02
Sync wasmtime and remove wasm32 deprecated module

kubkon:
Since `wasmtime` now uses `wasi` and `wasi32` modules, we can now
safely remove the `wasm32` module. This commit also updates `wasmtime`
to the latest upstream.


sunfishcode:
Thanks!

PR #174, opened 2019-11-06
Merge wasi-misc-tests repo as a subdir

kubkon:
This commit merges [CraneStation/wasi-misc-tests] repo as a subdir
of this repo while preserving **all** of git history. There is an
initiative to pull `wasi-common` and `wasi-misc-tests` into
[CraneStation/wasmtime], and [CraneStation/wasmtime] becoming a
monorepo. This came about for several reasons with a common theme
of convenience, namely, having a monorepo:
1. cleans up the problem of dependencies (as we have seen first
   hand with dependabot enabled, it can cause some grief)
2. completely removes the problem of syncing the closely dependent
   repos (e.g., updating wasi-common with say a bugfix generally
   implies creating a "sync" commit for pulling in the changes
   into the "parent" repo, in this case, `wasmtime`; similar case
   would happen for `wasi-misc-tests` vs `wasi-common` scenario)
3. mainly for the two reasons above, makes publishing to
   crates.io easier
4. hopefully streamlines the process of getting the community involved
   in contributing tests to `wasi-common` as now everything is one
   place

[CraneStation/wasi-misc-tests]: https://github.com/CraneStation/wasi-misc-tests
[CraneStation/wasmtime]: https://github.com/CraneStation/wasmtime

Additionally, while here, this commit also updates the dependencies and
cleans up `Cargo.toml` with details such as authors, etc.


wasi-misc-tests/LICENSE
```diff
@@ -0,0 +1,220 @@
+
```
marmistrz: Do we need a separate license here?

marmistrz (approved):
I'm definitely in favor of merging those two repos!

peterhuene (approved):
👍.  I also think we can forego the additional LICENSE file, but this looks great.  Thanks!

marmistrz:
@kubkon could we merge #116 first?

wasi-misc-tests/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "wasi-misc-tests"
+version = "0.1.0"
+authors = ["The Wasmtime Project Developers"]
+edition = "2018"
```
sunfishcode, line 5: We should add 'publish = false' here since this isn't meant to be published on crates.io in its current form.

wasi-misc-tests/LICENSE
sunfishcode: We'd need one if the crate would be published on crates.io, but I don't think we mean to do that for now, so we can omit the LICENSE if you want.

kubkon: Removed! We can re-add it later when and if we decide to publish `wasi-misc-tests` as a crate. :-)

kubkon:
> @kubkon could we merge #116 first?

Absolutely, that's the plan :-) In fact, I'll wait with this PR until #175 is also merged to avoid hellish rebases.

PR #175, opened 2019-11-06
Improvements to `WasiCtxBuilder`, and a couple bug fixes

acfoltzer:
Note that since this involves changes to the interface that downstream clients have to use, it requires a different commit of `wasmtime` for testing. That `wasmtime` commit is currently on my private fork, so I'm opening this as a draft, as it will need to be amended before merging (**edit:** see https://github.com/CraneStation/wasmtime/pull/490).

Text from the commit messages follows:

This changes the fields on the builder to types that let the various `.arg()`, `.env()`, etc methods
infallible, so we don't have to worry about handling any errors till we actually build. This reduces
line noise when using a builder in a downstream application.

Deferring the processing of the builder fields also has the advantage of eliminating the opening and
closing of `/dev/null` for the default stdio file descriptors unless they're actually used by the
resulting `WasiCtx`.

Now that failures are deferred until `WasiCtxBuilder::build()`, we don't need to have `Result` types
on the other methods any longer.

Additionally, using `IntoIterator` rather than `Iterator` as the trait bound for these methods is
slightly more general, and saves the client some typing.

Unicode errors when inheriting arguments and environment variables no longer cause a panic, but
instead go through `OsString`. We return `ENOTCAPABLE` at the end if there are NULs, or if UTF-8
conversion fails on Windows.

This also changes the bounds on some of the methods from `AsRef<str>` to `AsRef<[u8]>`. This
shouldn't break any existing code, but allows more flexibility when providing arguments. Depending
on the outcome of WebAssembly/WASI#8 we may eventually want to require
these bytes be UTF-8, so we might want to revisit this later (**edit:** we now validate UTF-8 during `build()`).

Finally, this fixes a tiny bug that could arise if we had exactly the maximum number of file
descriptors when populating the preopens, and a bug in the Linux implementation of `isatty()`.

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -1,19 +1,81 @@
 use crate::fdentry::FdEntry;
-use crate::sys::dev_null;
 use crate::{wasi, Error, Result};
 use std::borrow::Borrow;
 use std::collections::HashMap;
 use std::env;
-use std::ffi::CString;
+use std::ffi::{CString, OsString};
 use std::fs::File;
 use std::path::{Path, PathBuf};
 
+enum PendingFdEntry {
+    Thunk(fn() -> Result<FdEntry>),
+    File(File),
+}
+
+impl std::fmt::Debug for PendingFdEntry {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            PendingFdEntry::Thunk(f) => write!(
+                fmt,
+                "PendingFdEntry::Thunk({:p})",
+                f as *const fn() -> Result<FdEntry>
+            ),
+            PendingFdEntry::File(f) => write!(fmt, "PendingFdEntry::File({:?})", f),
+        }
+    }
+}
+
+#[derive(Debug, Eq, Hash, PartialEq)]
+enum PendingCString {
+    Bytes(Vec<u8>),
+    OsString(OsString),
+}
+
+impl From<Vec<u8>> for PendingCString {
+    fn from(bytes: Vec<u8>) -> Self {
+        Self::Bytes(bytes)
+    }
+}
+
+impl From<OsString> for PendingCString {
+    fn from(s: OsString) -> Self {
+        Self::OsString(s)
+    }
+}
+
+impl PendingCString {
+    #[cfg(not(windows))]
+    fn into_bytes(self) -> Result<Vec<u8>> {
+        use std::os::unix::ffi::OsStringExt;
+        match self {
+            PendingCString::Bytes(v) => Ok(v),
+            // on Unix, we use the bytes from the CString directly
+            PendingCString::OsString(s) => Ok(s.into_vec()),
+        }
+    }
+
+    #[cfg(windows)]
+    fn into_bytes(self) -> Result<Vec<u8>> {
+        match self {
+            PendingCString::Bytes(v) => Ok(v),
+            // on Windows, we go through conversion into a `String` in order to get bytes
+            PendingCString::OsString(s) => s.into_string().map(|s| s.into_bytes()),
+        }
+        .map_err(|_| Error::ENOTCAPABLE)
+    }
+
+    fn into_cstring(self) -> Result<CString> {
+        self.into_bytes()
+            .and_then(|v| CString::new(v).map_err(|_| Error::ENOTCAPABLE))
+    }
+}
+
 /// A builder allowing customizable construction of `WasiCtx` instances.
 pub struct WasiCtxBuilder {
-    fds: HashMap<wasi::__wasi_fd_t, FdEntry>,
+    fds: HashMap<wasi::__wasi_fd_t, PendingFdEntry>,
     preopens: Vec<(PathBuf, File)>,
-    args: Vec<CString>,
-    env: HashMap<CString, CString>,
+    args: Vec<PendingCString>,
+    env: HashMap<PendingCString, PendingCString>,
 }
 
 impl WasiCtxBuilder {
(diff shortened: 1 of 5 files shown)


acfoltzer:
I'm marking this as ready for review, since it doesn't look like https://github.com/CraneStation/wasmtime/pull/490 will pass until we have a final revision to use in wasmtime.

acfoltzer:
Per discussion with @sunfishcode, `WasiCtxBuilder::build()` now enforces that arguments and environment variables are valid UTF-8, including from `OsString`s inherited from the host process args/env.

src/ctx.rs
```diff
@@ -1,117 +1,185 @@
...
+    fn from(s: OsString) -> Self {
+        Self::OsString(s)
+    }
+}
+
+impl PendingCString {
+    fn into_string(self) -> Result<String> {
+        match self {
+            PendingCString::Bytes(v) => String::from_utf8(v).map_err(|_| Error::ENOTCAPABLE),
+            PendingCString::OsString(s) => s.into_string().map_err(|_| Error::ENOTCAPABLE),
```
sunfishcode: The error we use for character encoding issues is `EILSEQ`, so we should use that here too.

src/ctx.rs
```diff
@@ -1,117 +1,185 @@
...
+        }
+    }
+
+    /// Create a `CString` containing valid UTF-8, or fail with `Error::ENOTCAPABLE`.
+    fn into_utf8_cstring(self) -> Result<CString> {
+        self.into_string()
+            .and_then(|s| CString::new(s).map_err(|_| Error::ENOTCAPABLE))
```
sunfishcode: We use `EILSEQ` for reporting non-NUL-terminated strings too.

src/ctx.rs
```diff
@@ -1,117 +1,185 @@
...
+        builder
     }
 
     /// Add arguments to the command-line arguments list.
-    pub fn args<S: AsRef<str>>(mut self, args: impl Iterator<Item = S>) -> Result<Self> {
-        let args: Result<Vec<CString>> = args
-            .map(|arg| CString::new(arg.as_ref()).map_err(|_| Error::ENOTCAPABLE))
+    ///
+    /// Arguments must be valid UTF-8 with no NUL bytes, or else `WasiCtxBuilder::build()` will fail
+    /// with `Error::ENOTCAPABLE`.
```
sunfishcode: Update the comments to say `EILSEQ` here and elsewhere too :-).

src/ctx.rs
```diff
@@ -121,42 +189,74 @@ impl WasiCtxBuilder {
...
+            .env
+            .into_iter()
+            .map(|(k, v)| {
+                k.into_string().and_then(|mut pair| {
+                    v.into_string().and_then(|v| {
+                        pair.push('=');
+                        pair.push_str(v.as_str());
+                        // we have valid UTF-8, but the keys and values have not yet been checked
+                        // for NULs, so we do a final check here
+                        CString::new(pair).map_err(|_| Error::ENOTCAPABLE)
```
sunfishcode: Ditto, `EILSEQ`.

sunfishcode (commented):
Looks good! Deferring the error reporting until build time feels cleaner and more flexible.

acfoltzer:
Got it! I thought `ENOTCAPABLE` was a strange choice there but am glad to know there's a purpose-built errno for this.

src/ctx.rs
```diff
@@ -121,42 +189,73 @@ impl WasiCtxBuilder {
...
 
     /// Build a `WasiCtx`, consuming this `WasiCtxBuilder`.
-    pub fn build(mut self) -> Result<WasiCtx> {
-        // startup code starts looking at fd 3 for preopens
-        let mut preopen_fd = 3;
+    ///
+    /// If any of the arguments or environment variables in this builder cannot be converted into
+    /// `CString`s, either due to NUL bytes or Unicode conversions, this returns `Error::EILSEQ`.
+    pub fn build(self) -> Result<WasiCtx> {
+        // process arguments and environment variables into `CString`s, failing quickly if they
```
sunfishcode: Sorry for missing this the first time, but the Wasmtime comment style is to start sentences with capital letters. Here and in a few other places in the patch.

acfoltzer: Got 'em!

src/sys/unix/linux/mod.rs
```diff
@@ -9,9 +9,11 @@ pub(crate) mod fdentry_impl {
         use nix::errno::Errno;
 
         let res = libc::isatty(fd.as_raw_fd());
-        if res == 0 {
+        if res == 1 {
```
kubkon, line 12: Ah, that's an excellent catch, thanks! While we're here, would you mind applying the same fix for the BSD module [here]? I have made a mistake while verifying this in the man pages when fixing this before. It turns out I was looking up the man pages for a utility binary instead of a libc fn :-)

[here]: https://github.com/CraneStation/wasi-common/blob/master/src/sys/unix/bsd/mod.rs#L10

src/ctx.rs
```diff
@@ -121,42 +189,73 @@ impl WasiCtxBuilder {
...
+    /// If any of the arguments or environment variables in this builder cannot be converted into
+    /// `CString`s, either due to NUL bytes or Unicode conversions, this returns `Error::EILSEQ`.
+    pub fn build(self) -> Result<WasiCtx> {
+        // Process arguments and environment variables into `CString`s, failing quickly if they
+        // contain any NUL bytes, or if conversion from `OsString` fails.
+        let args = self
+            .args
+            .into_iter()
+            .map(|arg| arg.into_utf8_cstring())
+            .collect::<Result<Vec<CString>>>()?;
```
kubkon, line 202: This is not a nitpick, just an FYI that you could have omitted the concrete type stored inside the container here, i.e., `Result<Vec<_>>` :-)

src/ctx.rs
```diff
@@ -1,117 +1,185 @@
...
+            .into_iter()
             .map(|t| {
                 let (k, v) = t.borrow();
-                let k = CString::new(k.as_ref()).map_err(|_| Error::ENOTCAPABLE);
-                let v = CString::new(v.as_ref()).map_err(|_| Error::ENOTCAPABLE);
-                match (k, v) {
-                    (Ok(k), Ok(v)) => Ok((k, v)),
-                    _ => Err(Error::ENOTCAPABLE),
-                }
+                (k.as_ref().to_vec().into(), v.as_ref().to_vec().into())
```
kubkon, line 161: I like the cleanup here :-)

src/ctx.rs
```diff
@@ -175,12 +274,12 @@ impl WasiCtx {
...
+    pub fn new<S: AsRef<[u8]>>(args: impl IntoIterator<Item = S>) -> Result<Self> {
         WasiCtxBuilder::new()
-            .and_then(|ctx| ctx.args(args))
-            .and_then(|ctx| ctx.inherit_stdio())
-            .and_then(|ctx| ctx.inherit_env())
-            .and_then(|ctx| ctx.build())
+            .args(args)
+            .inherit_stdio()
+            .inherit_env()
+            .build()
```
kubkon, line 282: Yeah, from the code readability point of view, that's so much better. Well done!

kubkon (approved):
LGTM! If you could only fix the bug in the BSD usage of `isatty` that would be great! :-)

src/sys/unix/linux/mod.rs
kubkon: I've now gone ahead and applied your fix to the BSD impl as well as I'd really like to merge this PR so that we can make progress on #174, and I wouldn't want to merge #174 before this one, as rebasing might be a bit of a headache :-)

kubkon:
> I'm marking this as ready for review, since it doesn't look like [CraneStation/wasmtime#490](https://github.com/CraneStation/wasmtime/pull/490) will pass until we have a final revision to use in wasmtime.

Yeah, that's a real hassle, isn't it? For that very reason, I'll be trying my best to migrate `wasi-common` inside of `wasmtime` ASAP. I've already made some headway by incorporating tests in this repo in #174, and after it lands, I'll get my hands dirty in trying to migrate `wasi-common` over to `wasmtime`. Also, on that topic, if you could withhold from creating any new PRs until then, that'd be great. Having said that, if you find anything that's critical, then by all means, please feel free to open a new PR to this repo :-)

acfoltzer:
Sorry I wasn't on the right time zone to offer more feedback here; thank you for merging and good catch with BSD!

